PR #12, opened 2020-03-23
Typo fixed in README.md

biranchi2018:
Fixed the following typo error:
-print("Emotion: ", demography["dominant_race"])
+print("Race: ", demography["dominant_race"])


serengil:
thank you

PR #20, opened 2020-04-21
Prevent re-detection on 224

uriafranko:
Set img_244 variable once for all greyscale face detection actions


serengil:
I guess the variable should be img_224 instead of img_244 :)

uriafranko:
Haha Sorry typo mistake...
Fixed

serengil:
Thank you for this contribution. In this way, detectFace function will be called 3 times less.

PR #21, opened 2020-04-22
feature: add port argument of api.py

Jasonnor:
Greetings, such a good facial analysis framework.
I simply added port option of `api.py` to switch port, and default port is `5000`.

Usage:
```bash
python api.py --port 5566
```

It will be useful for building with other service, thanks!


serengil:

PR #49, opened 2020-06-18
fix: check first the type of the input_shape before get the shape. Al…

danielgo13:
I think found some errors when do verify functions:
1. The if-else statements when checking the model_name broken at the end, so it will always return error when you use model_name parameter
2. When I tried to use FbDeepFace model, turns out the input shape of the input_layer is a list ex: [ (None, 192, 192, 3) ] . So the code breaks when using the DeepFace.

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -196,7 +196,7 @@ def verify(img1_path, img2_path=''
 			print("Using FB DeepFace model backend", distance_metric,"distance.")
 			model = FbDeepFace.loadModel()
 		
-		if model_name == 'DeepID':
+		elif model_name == 'DeepID':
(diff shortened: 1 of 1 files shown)


serengil:

serengil:
When I build a model and print its first layer's input shape, then it returns: (None, 224, 224, 3) but you got [(None, 224, 224, 3)]. That's why, I need to fix the code as illustrated below. Your modification throws exception in my environment. I would like to inform you.

```
input_shape = custom_model.layers[0].input_shape

if type(input_shape) == list: # [(None, 224, 224, 3)]
	input_shape = input_shape[0][1:3]
else: # (None, 224, 224, 3)
	input_shape = input_shape[1:3]
```

PR #86, opened 2020-08-26
Provide the ability to override the default gdrive model weights urls

deepuvsdeepu:
Off late I'm getting the below error while trying to load the VGG face models on Google Colab notebooks. 

vgg_face_weights.h5 will be downloaded...
Access denied with the following error:

 	Too many users have viewed or downloaded this file recently. Please
	try accessing the file again later. If the file you are trying to
	access is particularly large or is shared with many people, it may
	take up to 24 hours to be able to view or download the file. If you
	still can't access a file after 24 hours, contact your domain
	administrator. 

As of now, the model weight url is hard coded and I feel it would be helpful to provide the ability to override the same.

### DIFF
--- a/deepface/basemodels/DeepID.py
+++ b/deepface/basemodels/DeepID.py
@@ -8,7 +8,7 @@ import zipfile
 
 #-------------------------------------
 
-def loadModel():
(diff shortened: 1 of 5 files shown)


serengil:

PR #133, opened 2020-11-26
Fix preprocess_face for age prediction

sugyan:
`detector_backend` was ignored only when running age prediction.
It is probably missed the change when in commit https://github.com/serengil/deepface/commit/b9abfb9c00020c2bb616e52a6b5d0447f66afd0d


serengil:
thank your for your contribution.

PR #134, opened 2020-11-28
add lightgbm to requirments

ma7555:
needed for ensemble model


serengil:

PR #169, opened 2021-01-15
Update functions.py

tommybilwij:
To load OpenCV haar cascade model in align_face function, otherwise eye_detector is undefined

### DIFF
--- a/deepface/commons/functions.py
+++ b/deepface/commons/functions.py
@@ -365,6 +365,7 @@ def alignment_procedure(img, left_eye, right_eye):
 def align_face(img, detector_backend = 'opencv'):
 	
 	home = str(Path.home())
+	initialize_detector(detector_backend)
 	
 	if (detector_backend == 'opencv') or (detector_backend == 'ssd'):
 		


serengil:

DeepFace interface has verify, find and analyze functions. initialize_detector is called in those 3 functions.

What if you would not call those functions? Those functions call functions.preprocess_face. Then Preprocess_Face function calls detect_face and align_face respectively. If initialize_detector function has not been called before, initialize_detector is going to be called in detect face.

```
	if not "face_detector" in globals():
		initialize_detector(detector_backend = detector_backend)
```

Your PR causes to initialize detectors again and again. As I mentioned above, you should check face detector exists in globals and then initialize it in align_face. Once face_detector initialized, then eye_detector is always initialized.

tommybilwij:
Thanks so much for your detailed explanation!

I think when I only want to preprocess the images using functions.preprocess_face(..), it prompts to run align_face(..) without initialising the global variables from initialize_detector(..) hence eye_detector is not initialised. That's why I think it's better to add that line somewhere in between to initialise the global variables.

Screen Shot 2021-01-15 at 5 52 05 pm

Screen Shot 2021-01-15 at 5 51 45 pm

serengil:
You have to call detect face before align face because one image could store more than one faces.

PR #211, opened 2021-03-29
Fix bounding box for dlib

olive380:
Change right --> top


serengil:

PR #272, opened 2021-07-01
Finish the TODO in function "preprocess_face"

Libo-Xu:
Resize the detected face image proportionally then add black pixels to resize it to target_size, which avoids the transformation on the detected faces resulting from resizing directly.

### DIFF
--- a/deepface/commons/functions.py
+++ b/deepface/commons/functions.py
@@ -124,9 +124,22 @@ def preprocess_face(img, target_size=(224, 224), grayscale = False, enforce_dete
 	if grayscale == True:
 		img = cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)
 
-	img = cv2.resize(img, target_size)
-	#TODO: resize causes transformation on base image, you should add black pixels to rezie it to target_size
+	# img = cv2.resize(img, target_size)
+	
+	# First resize the longer side to the target size
+	factor = target_size[0] / max(img.shape) # Suppose target_size[0] == target_size[1]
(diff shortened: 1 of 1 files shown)


serengil:
Thank you for your contribution. I'm going to merge it in the next couple of days after testing.

serengil:

I notice that "# Suppose target_size[0] == target_size[1]" as a comment. Here, DeepID expects (55, 47, 3) shaped inputs. I tested this snippet for that shaped and it worked. That causes any problem?

Libo-Xu:
Thanks for your review.

First, I thought the parameter _target_size_ could only be (224, 224) or (48, 48) so it doesn't matter whether to use `factor = target_size[0] / max(img.shape)`  or ` factor = target_size[1] / max(img.shape)`. 

But as you say that the _target_size_ could also be like (55, 47), which also worked under this code. I think this is because **img.shape[1] * (target_size[0] / max(img.shape))** happened to be smaller than **target_size[1]** (i.e. 47). 

Actually, there is an implicit prerequisite for adding black pixels, which is 

> img.shape[0] <= target_size[0] && img.shape[1] <= target_size[1]

, note that the _**img**_ here is the image before the padding process. 

So I think there is a more general way to make sure that the prerequisite is satisfied, which is 
```
factor_0 = target_size[0] / img.shape[0]
factor_1 = target_size[1] / img.shape[1]
factor = factor_0 if factor_0 < factor_1 else factor_1
```

serengil:
BTW, I already merged you PL.

PR #312, opened 2021-08-07
model summary and path

Xu-jianwen:


serengil:
It seems that it is not a must change. I will not merge it into the master branch.

PR #318, opened 2021-08-14
Dev

Xu-jianwen:


serengil:
What is the purpose of this PL? It's hard to understand what it is doing.

I see hard coded paths in the final commit. This is not acceptable. I will not merge it to the master branch as is.

PR #323, opened 2021-08-23
Configure models path via ENV variable

RoelofRuis:
Dear Sefik, thanks for this already great library!

This PR proposes to optionally set the path at which the `.deepface` directory will be created via the `DEEPFACE_HOME` env variable. The directory falls back to `Path.home()` if the ENV variable is unset.


serengil:

PR #367, opened 2021-11-10
Remove extra BGR2RGB call in RetinaFaceWrapper

bhky:
I suggest removing the extra BGR2RGB call.

While the RetinaFace model does expect RGB images, the corresponding conversion will be done in the RetinaFace preprocessing:
https://github.com/serengil/retinaface/blob/0c8d3ed8bd223c5753978a5e32549c25edff3316/retinaface/commons/preprocess.py#L45

The extra BGR2RGB call here will in fact make the final channel order wrong, because OpenCV does not know about the channel order, each call will simply swap the channels.

Thanks!


serengil:

PR #392, opened 2021-12-22
Modified functions.py so that it can read from https

KananMahammadli:
The current deepface version supports image loading from the path or URL that should start with exactly "data:image/" or directly numpy array, if this kind of example [image](https://images.unsplash.com/photo-1438761681033-6461ffad8d80?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxzZWFyY2h8MXx8cGVyc29ufGVufDB8fDB8fA%3D%3D&w=1000&q=80) URL is given to analyze, it will raise an error - "Confirm that 'URL' exists", since the URL doesn't start with "data:image/", base64_img variable is False, exact_image variable is also False; therefore it will check the elif statement in the load_image function inside the functions.py, since img variable's value is not a path, but URL, it will raise ValueError. In conclusion, I added an option to read from the URLs starting with HTTP and convert it to NumPy array so that it won't raise any error and make a proper prediction.

### DIFF
--- a/deepface/commons/functions.py
+++ b/deepface/commons/functions.py
@@ -4,6 +4,8 @@ import pandas as pd
 import cv2
 import base64
 from pathlib import Path
+from PIL import Image
+import requests
 
 from deepface.detectors import FaceDetector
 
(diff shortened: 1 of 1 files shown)


serengil:

PR #411, opened 2022-01-18
Enhanced cropping of detected faces using dlib

sotirismos:
There are edge cases where the bounding box is partially outside the image window (that explains the fact that some coordinates of the bounding box in dlib have negative values). In these cases, cropping the detected face based on the box coordinates either fails or returns odd results.

So, instead of this command

detected_face = img[top:bottom, left:right]

we should get the detected face with the following command

detected_face = img[max(0, top): min(bottom, img_height), max(0, left): min(right, img_width)]


serengil:

PR #436, opened 2022-03-12
update functions.py

laxmankusuma:
and img2_path != None this condition is not required because in and condition if type of img2_path is string then defiantly it is not None.

example:

path1 = 'xyz path'
type(path1)
output : str

path2 = None
type(path2)
output : NoneType


serengil:
I added this condition to avoid an error but cannot remember what it is. To swim in safe zone, I prefer not to remove it.

PR #440, opened 2022-03-15
Fix typo in analyze docstring

renhaa:
```python
models['race'] = DeepFace.build_model('race')
```
gives
``` 
ValueError: Invalid model_name passed - race
```
Should be capital "Race"
```python
models['race'] = DeepFace.build_model('Race')
```


serengil:

PR #447, opened 2022-03-23
Update README.md

nelson123-lab:
There is no mention of the weight for the model that we need to download when we are using the library for the first time. It is not mentioned in the installation part too.


serengil:
If you use vgg-face backend, it will a file size of 512 mb but if you use different model then size of the weight will change. We cannot use a general term here.

serengil:
Also, that is not going to be downloaded when you run pip install deepface.

PR #456, opened 2022-04-15
Allow silent operation for find()

eigenemara:
`DeepFace.find()` by default shows user warnings about already existing representations and function time elapsed, this commit allows silent operation by setting `silent=True` when calling `find()`


serengil:

I plan to manage progress bar argument with silent argument too

PR #471, opened 2022-05-08
Added command line interface via fire library

onuratakan:
Hello,

I am Onur, this project is awesome and i will use in my new and existing projects.

In this pull request i added CLI interface for DeepFace with this update users can be use DeepFace with this command
```console
DeepFace verify -IMG1_PATH ...
```

If you accept this idea i can write documentation for this update.

Best regards,

Onur Atakan ULUSOY

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -12,6 +12,8 @@ import pandas as pd
 from tqdm import tqdm
 import pickle
 
+import fire
+
 from deepface.basemodels import VGGFace, OpenFace, Facenet, Facenet512, FbDeepFace, DeepID, DlibWrapper, ArcFace, Boosting
 from deepface.extendedmodels import Age, Gender, Race, Emotion
 from deepface.commons import functions, realtime, distance as dst
@@ -822,3 +824,4 @@ def detectFace(img_path, target_size = (224, 224), detector_backend = 'opencv', 
 #main
 
 functions.initialize_folder()
+fire.Fire()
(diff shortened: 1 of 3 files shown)


serengil:

i plan to change the command to lowercase

    entry_points={
        "console_scripts":
        ["deepface = deepface:DeepFace"],
    }

serengil:
When i try to run the command, it is not working. Do you have an idea why this happens?

(sefik) sefik@Sefiks-MacBook-Pro deepface % deepface_cmd verify -IMG1_PATH tests/dataset/img1.jpg -IMG2_PATH tests/dataset/img2.jpg
Traceback (most recent call last):
  File "/Users/sefik/opt/anaconda3/envs/sefik/bin/deepface_cmd", line 33, in <module>
    sys.exit(load_entry_point('deepface', 'console_scripts', 'deepface_cmd')())
  File "/Users/sefik/opt/anaconda3/envs/sefik/bin/deepface_cmd", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/Users/sefik/opt/anaconda3/envs/sefik/lib/python3.8/importlib/metadata.py", line 79, in load
    return functools.reduce(getattr, attrs, module)
AttributeError: module 'deepface' has no attribute 'DeepFace'

serengil:
also, unit tests are not working anymore as is

serengil:
i have to remove fire based changes - unfortunately

onuratakan:
Okey, i will fixit. and re send a pr. Maybe we should add workflow based tests for errors =)

onuratakan:
The problem is if __main__ section when we import the DeepFace also we run the fire command (fire.fire()).

So we should add an  if __name__ == '__main__':

I can understand your revert operation, its necessary, but i want to fix and add to this amazing project =) let's go

(A litle crazy)

PR #474, opened 2022-05-09
Opencv SFace implementation

rodrigo2019:
Recently opencv has implemented SFace for recognition and YuNet for detection. These model have a good performance, in some cases SFace have better performance than Facenet.

[https://docs.opencv.org/4.x/d0/dd4/tutorial_dnn_face.html](https://docs.opencv.org/4.x/d0/dd4/tutorial_dnn_face.html)
[https://github.com/opencv/opencv_zoo/tree/master/models/face_recognition_sface](https://github.com/opencv/opencv_zoo/tree/master/models/face_recognition_sface)


serengil:
very nice work

we should fine a threshold for this model but it does not block to merge. i can do it later. thank your for your contribution

serengil:
i found the threshold values and updated commons/distance.py

PR #497, opened 2022-06-18
Feature/non binary gender result

shirasael:


serengil:
thank you for your contribution but it seems this is not the gender proba change.

1- in DeepFace.py resp_obj was dict but not it is list. 

2- in DeepFace.py, if you return gender_prediction directly, that would be the proba of genders. 

3- in ArcFace.py, why you are not using keras.models.Model anymore?

shirasael:
Hi,

You are correct - I originally opened this branch to change the behavior of the analyze function in DeepFace to return the gender probability, and ended up adding a few more changes.

Regarding you comments:

1 - I realize that, but why? It is much easier to access the results if it were a list (i.e `result[0]` instead of something like `result['instance_'+str(0]`). I noticed that the order is constant with the results, and there's an extra-processing at the end of the `analyze()` function to convert the results from list to dict. Personally I found it redundant and inconvenient, but I would love to hear your take on why this was the design :)

2 - True. I guess I just used the same process as was used in other analysis actions. I can change that to return the gender prediction directly (should I multiply by 100 though?)

3 - Good point, I am not sure why / when I made this change. Will revert.

serengil:
1- because, i realized that sometimes python returns array unsorted. i can understand the real index of that item with instance_x key. its initial version was a list instead of dict.

2- yes, scores should be in scale of [0, 100] similar to emotion and race scores.

shirasael:
1 - As far as I know lists maintain constant order (if this weren't true than the dict would also not be sorter, since it depends of the ondexes in the list)

2 - No problem, will make the change :)

serengil:
unsorted arrays might be a python related bug but i can access it in the dict with instance_101 independently from its order.

besides, would you please move the content of test_nonbinary_gender.py to unit_tests.py. because unit tests are running by github automations.

shirasael:
See this thread as referance:
https://stackoverflow.com/questions/13694034/is-a-python-list-guaranteed-to-have-its-elements-stay-in-the-order-they-are-inse

shirasael:
Sure, will move the tests 👍

PR #503, opened 2022-07-05
added cmake dependency

falkaabi:
Added cmake dependency to requirements.txt for those using python venv


.gitignore
```diff
@@ -20,3 +20,4 @@ tests/dataset/*.pkl
 .DS_Store
 deepface/.DS_Store
 *.pyc
+vDeepFace
```
falkaabi, line 23: This is the virtual environment I was using :)
No need to add it to the code

PR #563, opened 2022-09-24
Never print to stdout, always use stderr

Jille:
This makes it possible to use deepface in programs that use stdout for other usages. For example I use it in a streaming pipeline that outputs the represent results on stdout.

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -13,6 +13,7 @@ import uuid
 import json
 import time
 from tqdm import tqdm
+import sys
 
 #------------------------------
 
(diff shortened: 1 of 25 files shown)


Jille:
Keras still seems to print some stuff to stdout, I've yet to track down how to move that to stderr.

serengil:
those information are not error. they are printed to system out intendedly.

Jille:
stderr is for diagnostic messages, not necessarily errors.

https://stackoverflow.com/questions/3385201/confused-about-stdin-stdout-and-stderr has a lot of background

Andrea-Oliveri:
Good morning,
@Jille, if this is still an issue for you, you may want to use contextlib to redirect the stdout to stderr or to any other stream.

PR #577, opened 2022-10-21
Silent tensorflow models output to stdout

nguyenhoan1988:


serengil:

newer tf dumps progress bar and it is really annoying, i could not find a solution and recommended downgrade tf. it will help developers a lot.

i will validate the PR and merge soon. thank you.

PR #580, opened 2022-10-23
Added confidence to FaceDetector.detect_face and .detect_faces methods

Andrea-Oliveri:
Good morning,
I have implemented the return of confidence scores from the FaceDetector.detect_face and FaceDetector.detect_faces methods as a third parameter: (cropped_face, region, confidence).

I have also adapted other functions in this package using these methods such that they take into account the change.

In a documentation for FaceDetector exists, this may be a worthwhile note:
- MTCNN, Retinanet, Mediapipe and SSD have confidences in range [0, 1]
- OpenCV and DLib have confidence >=0 (not capped at 1, by design of the detectors)
- If a detector backend does not support returning the scores (not the case for now but may be in the future) it will return None as confidence score.

Finally, please note that:
- SSD already automatically discards faces with confidence <0.9
- RetinaFace already automatically discards faces with confidence <0.9
- RetinaFace already automatically discards faces with confidence <0.7
It may make sense, in order to allow end user to select its own detection thresholds, to decrease these default thresholds or adding a parameter to FaceDetector.detect_faces

Thank you for this powerful package.

### DIFF
--- a/deepface/commons/functions.py
+++ b/deepface/commons/functions.py
@@ -110,7 +110,7 @@ def detect_face(img, detector_backend = 'opencv', grayscale = False, enforce_det
 	face_detector = FaceDetector.build_model(detector_backend)
 
 	try:
-		detected_face, img_region = FaceDetector.detect_face(face_detector, detector_backend, img, align)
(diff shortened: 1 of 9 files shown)


serengil:

PR #581, opened 2022-10-23
Fixed minor bug: all regions now are (left, top, width, height)

Andrea-Oliveri:
Good Morning,
I have noticed that when the local variable img_region in .detect_faces methods are initialized as:
`img_region = [0, 0, img.shape[0], img.shape[1]`

This breaks the convention of the rest of the region variables in the package, which follow: (left, top, width, height).

I have therefore changed these for consistency. It should be noted that this will have absolutely zero impact on the functionality of the functions, as this initialized variable is always unused in the detector wrappers. The only case in which this can have an impact is when functions.detect_faces is used with "skip" detector_backend.

### DIFF
--- a/deepface/commons/functions.py
+++ b/deepface/commons/functions.py
@@ -95,7 +95,7 @@ def load_image(img):
 
 def detect_face(img, detector_backend = 'opencv', grayscale = False, enforce_detection = True, align = True):
 
-	img_region = [0, 0, img.shape[0], img.shape[1]]
+	img_region = [0, 0, img.shape[1], img.shape[0]]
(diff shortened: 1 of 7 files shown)


serengil:
what is its impact when detection is skipped?

Andrea-Oliveri:
When detection is skipped in `functions.detect_faces`, the last two tuple items are now inverted.
However, this representation is the consistent one with all other image regions returned by the package.

PS: future improvement may be to use namedtuples instead to completely eliminate this issue.

serengil:

PR #617, opened 2022-12-11
case handling analyze function with single actions

younver:


serengil:
thank you for contribution

younver:
thank you for creating such powerful library ^^

PR #633, opened 2023-01-26
[Bugfix] DeepFace.represent model parameter bugfix and completed parameters docstring

ronaldpereira:
# Description

This PR bugfixes the model parameter to `DeepFace.represent()` and added the complete parameter docstring.

## Current behavior

`DeepFace.represent()` ignores the `model` parameter as it is always overwritten in the code invariantly, making the `model` parameter useless.

```py
model = build_model(model_name)
```

## New behavior

`DeepFace.represent()` only uses the `model_name` parameter and build the `model` only if the `model` parameter wasn't sent (`model = None`).

```py
if model is None:
    model = build_model(model_name)
```

## Tests
- [x] Run tests/unit_test.py

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -502,7 +502,7 @@ def represent(img_path, model_name = 'VGG-Face', model = None, enforce_detection
 	Parameters:
 		img_path (string): exact image path. Alternatively, numpy array (BGR) or based64 encoded images could be passed.
 
-        model_name (string): VGG-Face, Facenet, Facenet512, OpenFace, DeepFace, DeepID, Dlib, ArcFace, SFace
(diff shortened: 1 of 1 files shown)


serengil:
thank you for your contribution. actually i retired the model input argument in represent function and other functions. but it seems i forgot to drop it in represent function.

ronaldpereira:
> thank you for your contribution. actually i retired the model input argument in represent function and other functions. but it seems i forgot to drop it in represent function.

That's ok. I've reverted the change and removed the `model` parameter, but the docstring creating for the other parameters is still necessary, correct?

serengil:
yes, other arguments are still valid. thank you again.

PR #648, opened 2023-01-29
refactor: use const for model feature labels

ekkolon:
# Description

*The purpose of this change is to make the code base more maintainable.*

This PR refactors the `analyze` function in the `DeepFace` module, such that:
- Feature labels for the *Gender*, *Emotion* and *Race* model are moved to and directly imported from their respective model namespace.

## Type of change

- [x] Refactoring (move/rename variables) 

## Current behavior
- Feature labels are declared within the *analyze* function, making it more difficult
to use those labels in (possible) future tests.

## Improvements
- Feature labels are declared and imported from the model they correspond to.
This makes it easier to add more features in the future.

__Assumption__
*Right now, space is allocated for each iteration for each list of feature labels in the `analyze` function.
We can expect a very small performance improvement when using constant values for the feature labels, since space is only allocated once. Since I haven't tested it, this remains an assumption.*

# Tests
- [x] Run tests/unit_test.py

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -26,6 +26,9 @@ from deepface.basemodels import (
     SFace,
 )
 from deepface.extendedmodels import Age, Gender, Race, Emotion
+from deepface.extendedmodels.Emotion import EMOTION_LABELS
+from deepface.extendedmodels.Gender import GENDER_LABELS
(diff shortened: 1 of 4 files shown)


serengil:
thank you for your contribution.

PR #660, opened 2023-02-06
Added HEIC support

tttdevelop:
OpenCV still does not support HEIC file format. This is a workaround to support HEIC files and convert to OpenCV format.

### DIFF
--- a/deepface/commons/functions.py
+++ b/deepface/commons/functions.py
@@ -3,6 +3,7 @@ import base64
 from pathlib import Path
 from PIL import Image
 import requests
+import pyheif
 
 # 3rd party dependencies
 import numpy as np
@@ -70,7 +71,20 @@ def load_image(img):
 
     # ---------------------------
 
-    if base64_img is True:
+    filename, file_extension = os.path.splitext(img)
+	if (file_extension == ".HEIC"):
+		heif_file = pyheif.read(img)
+		pil_image = Image.frombytes(
+			heif_file.mode, 
+			heif_file.size, 
+			heif_file.data,
+			"raw",
+			heif_file.mode,
+			heif_file.stride,
+			)
+		img = np.array(pil_image) 
+		img = img[:, :, ::-1].copy()         
+	elif base64_img is True:
         img = loadBase64Img(img)
 
     elif url_img is True:
--- a/deepface/commons/functions.py
+++ b/deepface/commons/functions.py
@@ -72,19 +72,20 @@ def load_image(img):
     # ---------------------------
 
     filename, file_extension = os.path.splitext(img)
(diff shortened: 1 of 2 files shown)


deepface/commons/functions.py
```diff
@@ -70,7 +71,21 @@ def load_image(img):
 
     # ---------------------------
 
-    if base64_img is True:
+    filename, file_extension = os.path.splitext(img)
```
serengil, line 74: this logic should be done in `exact_image is not True` block

requirements.txt
```diff
@@ -14,4 +14,5 @@ dlib>=19.20.0
 retina-face>=0.0.1
 mediapipe>=0.8.7.3
 fire>=0.4.0
-gunicorn>=20.1.0
\ No newline at end of file
+gunicorn>=20.1.0
+pyheif>=0.7.1
```
serengil, line 18: requirements.txt is just showing dependencies. we are not using this in the installation of the library. this dependency should be added in setup.py.

serengil:
thank you for your contribution.

i added some comments.

deepface/commons/functions.py
tttdevelop: No problems. Before I update the code, is this the format you are looking for ?

```
    elif exact_image is not True:  # image path passed as input
        if os.path.isfile(img) is not True:
            raise ValueError(f"Confirm that {img} exists")

        if (file_extension.lower() == ".heic"):
            ...
        else:
            img = cv2.imread(img)

```

bigcat88:
It may be easier to use [pillow-heif](https://github.com/bigcat88/pillow_heif) for this (code for the latest version 0.10.0):

``` python3
          img = np.array(pillow_heif.open_heif(img, bgr_mode=true))
```
Or use even a light version called pi-heif, because it seems that you only need decoding functionality.

tttdevelop:
Interesting, I'll look at both, maybe pi-heif is a better option.

Vincent-Stragier:
@tttdevelop,

[pi-heif](https://pypi.org/project/pi-heif/) looks powerful. Apparently, you can use it as a Pillow plugin. From the doc:

```python
from PIL import Image
from pi_heif import register_heif_opener

register_heif_opener()

im = Image.open("images/input.heic")  # do whatever need with a Pillow image
im.show()
```

So only `pi-heif` is required and two changes in the code:
```python
from pi_heif import register_heif_opener
register_heif_opener()
```

tttdevelop:
Thanks for the tip. This fell off my radar a bit so thanks also @Vincent-Stragier for shaking this up. I'll update with the pi-heif.

serengil:
closed due to inactivity

PR #681, opened 2023-02-25
cleaned gitignore

younver:
replaced all separate `__pycache__` and `.DS_Store` ignores with single ones to make `.gitignore` cleaner


serengil:

younver:
it's a pleasure~~

PR #686, opened 2023-02-28
edit Dockerfile

KhanbalaRashidov:
WORKDIR /app creates a new folder when you write. mkdir /app does not need to write


serengil:
working fine as is.

PR #694, opened 2023-03-09
Few Fixes and Features

ygorluizfrazao:
Added the option to set enforce_detection variable via POST method. Fixed issue when returning an error state, the system was messed up not knowing if the result was a tuple or dictionary.

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -1,9 +1,304 @@
-import argparse
-import app
+import warnings
+warnings.filterwarnings("ignore")
 
-if __name__ == "__main__":
-    deepface_app = app.create_app()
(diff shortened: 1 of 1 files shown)


serengil:
currently, deepface api is serving via gunicorn. you break this because you copied app.py functions to api.py directly. app should be run from api.py. 

`deepface_app = app.create_app()`

ygorluizfrazao:
Thanks for the answer, I'll check what I have done wrong.

This version base is from around 2 months ago, It happens that I had just
created the pull request now.

Hope success,

Em qui., 9 de mar. de 2023 12:35, Sefik Ilkin Serengil <
***@***.***> escreveu:

PR #696, opened 2023-03-09
Add Face Confidence to Deepface.represent()

liechtym:
For my use case, I need to be able to filter representation embeddings by the face detection confidence. In some cases, detectors are giving false positives, so I need to be able to tweak the face confidence threshold in my analysis. With the current implementation, I would have to do face detection and face representation in separate steps to get the confidence.


serengil:
totally makes sense. thank you for your contribution. i will merge it after pipeline worked successfully.

liechtym:
@serengil Thank you! Glad to contribute.

PR #700, opened 2023-03-15
Update README.md

nelson123-lab:
This library was very useful for me when I worked as a data labeling intern. Making face data of celebrities. The noisy images were easily removed with the help of the deepFace libraries' gender detection capability.


serengil:
ty but this should not be added to read me

Issue #435, opened 2022-03-10
DeepFace.verify given different results with same config and same image


thelostpeace:
**code snippets**

```
class FaceVerificationModel:
    def __init__(self):
        self.model = DeepFace.build_model('VGG-Face')
        self.detector_backend = 'opencv'
        self.enforce_detection = False
        self.distance_metric = 'euclidean_l2'
        self.prog_bar = False

    def preprocess(self, req):
        images = []
        for image in req.images:
            img = np.asarray(bytearray(image), dtype=np.uint8)
            img = cv2.imdecode(img, flags=cv2.IMREAD_COLOR)
            #img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
            images.append(img)
        custom = json.loads(req.custom_data)
        compare_idx = custom['compare_index']
        
        return [images[idx] for idx in range(len(images)) if idx != compare_idx], images[compare_idx]

    def process(self, req):
        cand_images, comp_image = self.preprocess(req)
        img_list = [[comp_image, cand] for cand in cand_images]
        result = DeepFace.verify(img1_path=img_list, img2_path=None, enforce_detection=self.enforce_detection, distance_metric=self.distance_metric, model=self.model, detector_backend=self.detector_backend, prog_bar=self.prog_bar)
        keys = [f'pair_{i + 1}' for i in range(len(img_list))]
        results = []
        for key in keys:
            results.append(result[key])

        return results
```

**dependencies and version**

 - python: 3.8.12
 - tensorflow: 2.8.0
 - deepface: 0.0.73

I am running deepface with CPU. There are my test images:
![1](https://user-images.githubusercontent.com/48209530/157626872-aa1b3bdb-6604-4b16-ac9f-92c5e0305ecf.jpeg)
![1](https://user-images.githubusercontent.com/48209530/157627000-0d043161-7884-4ea4-910a-f6e568ea5585.jpeg)

**test results**

```
{"verified": false, "distance": 0.9320651140425925, "threshold": 0.86, "model": "VGG-Face", "detector_backend": "opencv", "similarity_metric": "euclidean_l2"}
{"verified": false, "distance": 0.901325783194036, "threshold": 0.86, "model": "VGG-Face", "detector_backend": "opencv", "similarity_metric": "euclidean_l2"}
{"verified": false, "distance": 0.9320651140425925, "threshold": 0.86, "model": "VGG-Face", "detector_backend": "opencv", "similarity_metric": "euclidean_l2"}
[...]

PR #728, opened 2023-04-21
issue 435, measure eyes' size

DonghwanKIM0101:
@thelostpeace comments that the data is [x, y, w, h], not [x1, y1, x2, y2].

https://github.com/serengil/deepface/issues/435


serengil:

PR #738, opened 2023-04-28
Add extension list

PaulSamuelsonSI:


PaulSamuelsonSI:
I have a use case with jpg, jpeg and png files, with the same base filenames, stored in the same folder. The jpg is full resolution, the jpeg is a thumbnail and the PNG is transparent with the background removed. Because all the images contain the same person, when I use DeepFace.find() I do not want to search all 3 file types. This pull request adds an extension_list option to the find(), with the default being the original list, for backward compatibility.

PaulSamuelsonSI:
This was my first-ever pull request. I see it was closed, not merged. Did I do something wrong, or did you just choose not to implement it? Thanks.

PR #786, opened 2023-06-23
Fix potential bug in OpenCvWrapper

Vincent-Stragier:
Hello,

Here are some change to prevent some bug in `OpenCvWrapper.py`:

1. `score` is being declared in a `try:except:` block, thus it might not exist and raise an error later on.
2. No exception is specified in the `try:except:` block, which means the keyboard interrupt may be ignored.
3. Last but not least, checking for the size of the `faces` list before using it in a for loop is probably useless.

Best,
Vincent

### DIFF
--- a/deepface/detectors/OpenCvWrapper.py
+++ b/deepface/detectors/OpenCvWrapper.py
@@ -44,7 +44,9 @@ def detect_face(detector, img, align=True):
     detected_face = None
     img_region = [0, 0, img.shape[1], img.shape[0]]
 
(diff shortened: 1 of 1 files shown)


serengil:
if exception is thrown in try except block then length of faces will be 0 and scores will not be used. resp will be none in this case. parent program checks result is none or not.

that is why, it is better not to merge this PR.

Vincent-Stragier:
I still think number two should be addressed somehow. See https://www.flake8rules.com/rules/E722.html.

PR #801, opened 2023-07-13
Fix yunet return negative coordinates

AnthraX1:
Yunet returns negative coordinates if it thinks part of the detected face is outside the frame. This fixes that by reset the start point to zero.


serengil:
pylint fails. would you please fix them.

```
************* Module deepface.detectors.YunetWrapper
deepface/detectors/YunetWrapper.py:58:8: R1731: Consider using 'x = max(x, 0)' instead of unnecessary if block (consider-using-max-builtin)
deepface/detectors/YunetWrapper.py:60:8: R1731: Consider using 'y = max(y, 0)' instead of unnecessary if block (consider-using-max-builtin)
```

serengil:
also, adding a comment would be good in that line why we did it.

AnthraX1:
> also, adding a comment would be good in that line why we did it.

done

AnthraX1:
> This time fails because of this:
> 
> deepface/detectors/YunetWrapper.py:59:0: C0301: Line too long (105/100) (line-too-long)

It was complaining about the wrong line but the new commit should fix it.

serengil:
************* Module deepface.detectors.YunetWrapper
deepface/detectors/YunetWrapper.py:61:65: C0303: Trailing whitespace (trailing-whitespace)
deepface/detectors/YunetWrapper.py:47:8: W0105: String statement has no effect (pointless-string-statement)

AnthraX1:
finally passed all checks...

PR #805, opened 2023-07-19
Function analyze doesn't return an error if img_path is NoneType

DavidSheltonSF:
When I passed a wrong path to cv2.imread(), the function returned a NoneType, so when that return was passed to DeepFace.analyze() the function tried to continue processing and this genered more than one error.

So I added a condition in the function analyze() to check if img_path is NoneType. If it is NoneType, the function returns TypeError.
image

That was my main goal, however to make the code work on my computer I needed to make some other changes:

1º To run unit_tests.py I needed to get the absolute path to the images in the dataset directory and to the dataset directory itself, otherwise it would generate an error when the code tried to call functions, like analyze() for example. The print on the last line was for debugging.
image

2º The imports for version 2 of tensorflow didn't work for me and VScode showed a red line, so I had to use the imports for version 1 to get the code to work. Changing the Gender.py and VGGFace.py file imports was enough.
image
image

Infelizmente precisei fazer mais alterações do que gostaria, não sei se as versões do python e do sistema operacional que uso foram o motivo de tudo isso.  

Unfortunally, I needed to change more than I wanted to. I don't know if my python version or my OS version was the cause of all this.

Python version 3.8.10
OS Win7 64-bit

### DIFF
--- a/deepface/basemodels/VGGFace.py
+++ b/deepface/basemodels/VGGFace.py
@@ -18,8 +18,17 @@ if tf_version == 1:
         Activation,
     )
 else:
-    from tensorflow.keras.models import Model, Sequential
-    from tensorflow.keras.layers import (
+    # from tensorflow.keras.models import Model, Sequential
+    # from tensorflow.keras.layers import (
+    #     Convolution2D,
+    #     ZeroPadding2D,
+    #     MaxPooling2D,
+    #     Flatten,
+    #     Dropout,
+    #     Activation,
+    # )
+    from keras.models import Model, Sequential
+    from keras.layers import (
         Convolution2D,
         ZeroPadding2D,
         MaxPooling2D,
--- a/deepface/commons/functions.py
+++ b/deepface/commons/functions.py
@@ -25,7 +25,7 @@ if tf_major_version == 1:
     from keras.preprocessing import image
 elif tf_major_version == 2:
     from tensorflow.keras.preprocessing import image
-
+    
 # --------------------------------------------------
 
 
--- a/deepface/extendedmodels/Gender.py
+++ b/deepface/extendedmodels/Gender.py
@@ -9,14 +9,18 @@ from deepface.commons import functions
 # -------------------------------------
 # dependency configurations
 
+# A minha versão do tensorflow é a 2
 tf_version = int(tf.__version__.split(".", maxsplit=1)[0])
+print(tf_version)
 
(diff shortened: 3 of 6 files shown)


serengil:
the reason why vs code warns you for tf 2 imports, you do not have tf2.X dependency. if you remove those imports, then tf2 users will not be able to use deepface anymore!

deepface/extendedmodels/Gender.py
```diff
@@ -9,14 +9,17 @@
...
 
 if tf_version == 1:
     from keras.models import Model, Sequential
     from keras.layers import Convolution2D, Flatten, Activation
 elif tf_version == 2:
-    from tensorflow.keras.models import Model, Sequential
-    from tensorflow.keras.layers import Convolution2D, Flatten, Activation
+    # from tensorflow.keras.models import Model, Sequential
+    # from tensorflow.keras.layers import Convolution2D, Flatten, Activation
+    from keras.models import Model, Sequential
```
serengil, line 21: if you delete these dependencies, then tf2 users will not be able to use deepface anymore. the imports you just added if tf1 as seen.

deepface/basemodels/VGGFace.py
```diff
@@ -18,8 +18,17 @@
...
+    # from tensorflow.keras.models import Model, Sequential
+    # from tensorflow.keras.layers import (
+    #     Convolution2D,
+    #     ZeroPadding2D,
+    #     MaxPooling2D,
+    #     Flatten,
+    #     Dropout,
+    #     Activation,
+    # )
+    from keras.models import Model, Sequential
```
serengil, line 30: you make deepface unusable for tf2 users!

deepface/commons/functions.py
```diff
@@ -25,7 +25,7 @@
     from keras.preprocessing import image
 elif tf_major_version == 2:
     from tensorflow.keras.preprocessing import image
-
+
```
serengil, line 28: no logical change is available in the file. restore it please.

serengil (commented):
review cycle 1 comments

serengil:
Changes in unit test file will break github actions: https://github.com/serengil/deepface/blob/master/.github/workflows/tests.yml#L44

please do not do anything in unit tests except adding new scenarios

DavidSheltonSF:
Got it, I believe this pull request is not necessary. Thanks for the feedback.

Issue #697, opened 2023-03-13
Keras models have a memory leak when used in a loop


ethanrwilson1998:
I was having a creeping memory issue calling DeepFace.represent(...) in my training loop with Facenet512.  System memory was increasing by just a few MB per batch, but after a few hundred iterations my model's training would slow exponentially.  I narrowed it down to the model.predict call and found that it was using a Keras backend.  Replacing the model.predict(img) call with model(img) stopped the leak but was slow.  I am on tensorflow 2.11.0 and Keras 2.11.0, which I believe are currently the default versions when installing this package with pip.

I found this post (https://github.com/keras-team/keras/issues/13118) and followed some of the solutions.  I placed the following code at the top of my train code: 
```
import keras.backend as K
cfg = K.tf.compat.v1.ConfigProto()
cfg.gpu_options.allow_growth = True
K.set_session(K.tf.compat.v1.Session(config=cfg))
```
and periodically (every 1024 iterations) will call
```
K.clear_session()
```
This fixed my problem.

@serengil I don't know if it would make sense to put the solution I came up with into the package because calling keras.backend could affect peoples' code if they are using their own Keras models.  But when looking at old issues to try to find the problem I was facing, I saw that a few people have brought up issues running deepface in a loop before.  I think that this is the cause.

serengil:
Good workaround. But IMO this should be called in client side not in the library. Closing this issue. Hope this helps people having same problem.

serengil:
or i can check how many times represent function called in the current session, and apply this if it is greater than a threshold value.

LoganRebecca:
same here, I try to test with really simple code:
import torch
from tqdm import tqdm
from deepface import DeepFace
# Model
img = '6.jpg'  # or file, PIL, OpenCV, numpy, multiple
import tensorflow.keras.backend as K
# Inference
for i in tqdm(range(1000)):
        K.clear_session()
        objs = DeepFace.analyze(img_path = img,
                actions = ['age', 'gender', 'race'],detector_backend ='dlib'
                )
        print(objs)
        import gc
        gc.collect()

And the memory leak is there until OOM. can you help me solve this issue?

qcbegin:
I encountered the same memory problem when iteratively calling `analyze()`. I checked the post mentioned by @ethanrwilson1998 and they mentioned that substituting the `predict()` with `predict_on_batch()` would perfectly solve this issue:

[...]

PR #857, opened 2023-10-10
change model.predict() to model(), notably improve memory leak (issue #697)

Villyli:
To help improve #697, I use `model()` replace `model.predict()` and get notably improvement, which aviod generates a data generator in  `model.predict()`.

In my case, for a 1000 times loop, the memory increment is about 50 MiB. it seems that `DeepFace.represent()` now is capable of 100k loop for most use.

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -693,7 +693,7 @@ def represent(
         # represent
         if "keras" in str(type(model)):
             # new tf versions show progress bar and it is annoying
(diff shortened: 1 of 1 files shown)


Villyli:
Besides, u can use a redirect stdout warpper to clean annoying process output
```python
original_stdout = sys.stdout
sys.stdout = open(os.devnull, 'w')

# your codes intended to ignore output in terminal

sys.stdout.close()
sys.stdout = original_stdout
```

serengil:

PR #879, opened 2023-10-30
fix the problem of docker script which failed to run

yangsijie666:


serengil:
i will not approve this because dockerfile is working as is

Vmadmax:
Hi @serengil,
I can't run the script like this either. The path and image name are not correct. @yangsijie666 has corrected the errors.

For example:
**Dockerfile**
If you run the script dockerize.sh in the folder "scripts", the "Dockerfile" is needed to be in the same folder, because you run "docker build". But this file is in the root folder.
https://github.com/serengil/deepface/pull/879/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R298

**Image-Name**
The second issue is the image name. You tagged the image with the name "deepface_image", but you call it in the run command with "deepface".
For example wrong image name: https://github.com/serengil/deepface/pull/879/files#diff-38464d794cfc7bae2c6d50ec0e067c6c749f3ab7a1961827d1d34b5c49e26631R11

serengil:
@Vmadmax add the following line at the beginning of the script then

`cd ..`

serengil:
@Vmadmax 

Updating sh now

PR #896, opened 2023-11-28
Threshold Parameter Added

DataRohit:
**Threshold Argument Added**

**I appreciate the awesome deep face package! 🚀 While using it for my project, I found that the hardcoded threshold values in distances.py didn't suit my needs. For my contribution, I've added a parameter to make the threshold customizable. This marks my first-ever open-source contribution.**

 * `DeepFace.py`
   * `verify` function
   * `find` function

**Thanks,**
**Rohit Ingole (DataRohit)**

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -94,6 +94,7 @@ def verify(
     enforce_detection=True,
     align=True,
     normalization="base",
+    threshold="auto"
 ):
     """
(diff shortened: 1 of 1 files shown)


serengil:
Verify returns distance to make threshold customizable. If smo needs his/her own threshold, he/she should use distance in the response instead of verified key.

PR #900, opened 2023-11-29
Catch exceptions when decoding invalid images

### LINKED ISSUE #890: AttributeError: 'NoneType' object has no attribute 'shape'
AttributeError: 'NoneType' object has no attribute 'shape'

```
$ df = DeepFace.find(img_path="resources/face.jpeg", db_path="resources/dir_001", enforce_detection=False)

Finding representations:   0%|▎                                                                                                | 247/77736 [00:51<4:28:42,  4.81it/s]
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 df = DeepFace.find(img_path="resources/face.jpeg", db_path="resources/dir_001", enforce_detection=False)

File ~/.pyenv/versions/3.11.3/envs/cse/lib/python3.11/site-packages/deepface/DeepFace.py:488, in find(img_path, db_path, model_name, distance_metric, enforce_detection, detector_backend, align, normalization, silent)
    485 for index in pbar:
    486     employee = employees[index]
--> 488     img_objs = functions.extract_faces(
    489         img=employee,
    490         target_size=target_size,
    491         detector_backend=detector_backend,
    492         grayscale=False,
    493         enforce_detection=enforce_detection,
    494         align=align,
    495     )
    497     for img_content, _, _ in img_objs:
    498         embedding_obj = represent(
    499             img_path=img_content,
    500             model_name=model_name,
   (...)
    504             normalization=normalization,
    505         )

File ~/.pyenv/versions/3.11.3/envs/cse/lib/python3.11/site-packages/deepface/commons/functions.py:105, in extract_faces(img, target_size, detector_backend, grayscale, enforce_detection, align)
    103 # img might be path, base64 or numpy array. Convert it to numpy whatever it is.
    104 img = load_image(img)
--> 105 img_region = [0, 0, img.shape[1], img.shape[0]]
    107 if detector_backend == "skip":
    108     face_objs = [(img, img_region, 0)]

AttributeError: 'NoneType' object has no attribute 'shape'
```

I'm guessing cv2.imdecode can return None:
https://github.com/serengil/deepface/blob/99f5295638038cf554149edad450e4141583c807/deepface/commons/functions.py#L112

And then when we try to index that img object it throws the attribute error

ekkolon:
Previously, attempting to decode images using `load_image` failed without explicitly 
stating where and why the operation failed.

For example, we use `np.frombuffer` to interpret the image data as an 1-dimensional array.
This operation _DOES NOT_ fail when the image file is empty.

In contrast, passing invalid or empty image data to `cv2.imdecode` _will_ raise an execption.
Which is what happens when we don't check for a buffer size > 0.

This PR resolves these issues mentioned above by checking if the computed `chunk_arr` is empty,
and if so raising a legible `ValueError`.

Finally, we assume that if the `cv2.imdecode` fails it's because the image file's data is corrupt 
in one way or another and raise `ValueError` for that too.

Fixes #890

### DIFF
--- a/deepface/commons/functions.py
+++ b/deepface/commons/functions.py
@@ -117,11 +117,11 @@ def load_image(img):
             img = cv2.imdecode(chunk_arr, cv2.IMREAD_COLOR)
             return img
 
-        except cv2.error:
-            raise ValueError(f"Image '{img}' contains invalid data")
+        except cv2.error as exc:
+            raise ValueError(f"Image '{img}' contains invalid data") from exc
 
-        except Exception as e:
-            raise e
+        except Exception as exc:
+            raise exc
 
     # This causes troubles when reading files with non english names
     # return cv2.imread(img)
--- a/deepface/commons/functions.py
+++ b/deepface/commons/functions.py
@@ -107,10 +107,21 @@ def load_image(img):
 
     # For reading images with unicode names
     with open(img, "rb") as img_f:
-        chunk = img_f.read()
-        chunk_arr = np.frombuffer(chunk, dtype=np.uint8)
-        img = cv2.imdecode(chunk_arr, cv2.IMREAD_COLOR)
-    return img
+        try:
+            chunk = img_f.read()
+            chunk_arr = np.frombuffer(chunk, dtype=np.uint8)
+
+            if chunk_arr.size == 0:
+                raise ValueError(f"Image '{img}' is empty")
+
+            img = cv2.imdecode(chunk_arr, cv2.IMREAD_COLOR)
+            return img
+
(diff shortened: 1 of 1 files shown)


serengil:
`cv2.imread(img)` can do this but we gave up to use it because it fails when non-english characters are available. maybe, we should continue to use it and raise a meaningful message for non-english characters. will decide this soon.

serengil:
With PR - https://github.com/serengil/deepface/pull/911 - we started to use imread again and throw exception if file name has non english characters.

PR #901, opened 2023-11-30
Change BGR to RGB in docstrings

Rob1in:
Docstrings indicate that it expects BGR when it expects RGB.


serengil:
No, it expects how opencv loads which is BGR.

Issue #654, opened 2023-01-31
[Feature Request] Use (custom) Logger instead of print


ekkolon:
It would be of great use, if DeepFace used a more tailored logging solution/format other than print to report progress. Perhaps a small module that builds on top of the built-in logging module would suffice. 

Though using print reports progress/errors to library consumers, they can't opt in or out of it - and after all, the print message/error is all they get. 

Also, the following scenarios would be improved by such a feature:
- custom deployments
- custom log messages that use 'DeepFace:: [MESSAGE]' as namespace. 
- sending logs to custom location (when using more sophisticated architectures)
- overall, a more intuitive way to detect whats going on in the console, when using color schemes 
for the default log levels (DEBUG, INFO, WARNING, ERROR)

I'd like to hear your thoughts on this.

serengil:
thank you but i am very satisfied with the current printing approach

ekkolon:
I see you've re-opened this issue. Have you changed your mind about this?

Admittedly, reading through my initial suggestion, it sounds a bit shallow. 
It's definitely not a top priority feature, but would definitely enhance
integration and the developer experience.

I have no clear-cut concept about how this feature could be integrated into
DeepFace' workflow though. We could lay something out together, if you like.

serengil:
i closed it accidentally. currently, i am satisfied with the current printing but this does not mean it is not going to be changed. will think about this when i have time. that is why, labeled it as enhancement.

ekkolon:
All right. Sounds good to me!

PR #907, opened 2023-12-06
feat-task-654-use-custom-logger

serengil:
## Tickets

https://github.com/serengil/deepface/issues/654

## What has been done

DeepFace was using print in may locations which is pretty amateur. With this PR, this is switched to a logger class. Even though current logger solution prints everything in the console, now it is easy to customize.

## How to test

```shell
make test && make lint
```

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -470,8 +470,8 @@ def find(
         if not silent:
             logger.warn(
                 f"Representations for images in {db_path} folder were previously stored"
-                + f" in {file_name}. If you added new instances after the creation, then please "
-                + "delete this file and call find function again. It will create it again."
+                f" in {file_name}. If you added new instances after the creation, then please "
+                "delete this file and call find function again. It will create it again."
             )
 
         with open(f"{db_path}/{file_name}", "rb") as f:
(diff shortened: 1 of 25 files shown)


deepface/DeepFace.py
```diff
@@ -869,7 +876,7 @@ def detectFace(
             detected and aligned face as numpy array
 
     """
-    print("⚠️ Function detectFace is deprecated. Use extract_faces instead.")
+    logger.info("⚠️ Function detectFace is deprecated. Use extract_faces instead.")
```
serengil, line 879: we should use logger.debug instead of info and get rid of the emoji here

deepface/DeepFace.py
```diff
@@ -461,8 +468,8 @@ def find(
     if path.exists(db_path + "/" + file_name):
 
         if not silent:
-            print(
-                f"WARNING: Representations for images in {db_path} folder were previously stored"
+            logger.warn(
+                f"Representations for images in {db_path} folder were previously stored"
                 + f" in {file_name}. If you added new instances after the creation, then please "
```
serengil, line 473: no need to use plus here anymore

PR #909, opened 2023-12-06
Adding optional limit parameter to face analysis

### LINKED ISSUE #908: Please allow limiting number of faces for analysis
`DeepFace.analyze(...)` will process all the faces from an image by default, I'd like to limit the amount of time processing by providing a limit on number of faces to process. Faces should be ordered by size in pixels descending for this purpose.

## Background
Some of the images I've tried with had as many as five extra faces detected that were less than 100x100 pixels in an image with a 1430x1430 face in it, when I hadn't noticed the background persons before trying `DeepFace.analyze` on the image.

I'd like to use this in an automated fashion without needing to pre-screen images for crowds to ensure I don't have my process stuck on processing irrelevant faces for an inordinate amount of time.

sanguivore-easyco:
Resolves #908 

The old behavior is retained if the new parameter is not specified.


serengil:
I do not believe this is solving a common issue.

PR #917, opened 2023-12-14
fix(represent): the `represent` is broken when user-given image. Fix it and support `pathlib.Path` for img_path

fseasy:
# Background

I want to directly use the `represent` function on pre-detected face images. 

When I just pass the image path in pathlib.Path, it raise error on not-supported type. But I think it should support it because it's popular in Python3. OK, it's a tiny problem, and I just change the image path to str.

After given the image path in str, It still failed. Because 

https://github.com/serengil/deepface/blob/master/deepface/DeepFace.py#L706

return a tuple of (img, image_name),  but here we assign wrong.

What's more, the `region` is also wrong type: It should be a dict of x,y,w,h if from `extract_faces`, but here is a list.

## Changes

I fixed them all, in detail:

1. support `represent` user given image 
2. img_path support pathlib.Path

And test is added on `represent` user given image. Docs for `reprent` on return types is also updated.

Please give me some advice if it's not ok.

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -674,16 +674,30 @@ def represent(
             This might be convenient for low resolution images.
 
             detector_backend (string): set face detector backend to opencv, retinaface, mtcnn, ssd,
-            dlib, mediapipe or yolov8.
+            dlib, mediapipe or yolov8. A special value `skip` could be used to skip face-detection
+            and only encode the given image.
 
             align (boolean): alignment according to the eye positions.
 
             normalization (string): normalize the input image before feeding to model
 
     Returns:
-            Represent function returns a list of object with multidimensional vector (embedding).
-            The number of dimensions is changing based on the reference model.
-            E.g. FaceNet returns 128 dimensional vector; VGG-Face returns 2622 dimensional vector.
+            Represent function returns a list of object, each object has fields as follows:
+            {
+                // Multidimensional vector
+                // The number of dimensions is changing based on the reference model.
+                // E.g. FaceNet returns 128 dimensional vector; VGG-Face returns 2622 dimensional vector.
+                "embedding": np.array, 
+                
+                // Detected Facial-Area by Face detection in dict format.
+                // (x, y) is left-corner point, and (w, h) is the width and height
+                // If `detector_backend` == `skip`, it is the full image area and nonsense.
+                "facial_area": dict{"x": int, "y": int, "w": int, "h": int},
+                
+                // Face detection confidence. 
+                // If `detector_backend` == `skip`, will be 0 and nonsense.
+                "face_confidence": float
+            }
     """
     resp_objs = []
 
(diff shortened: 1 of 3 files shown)


deepface/DeepFace.py
```diff
@@ -702,23 +716,23 @@ def represent(
...
+            # Try load. If load error, will raise exception internal
+            img, _ = functions.load_image(img_path)
         # --------------------------------
         if len(img.shape) == 4:
             img = img[0]  # e.g. (1, 224, 224, 3) to (224, 224, 3)
         if len(img.shape) == 3:
             img = cv2.resize(img, target_size)
-            img = np.expand_dims(img, axis=0)
-            # when represent is called from verify, this is already normalized
+            img = np.expand_dims(img, axis=0) # Why we remove a axis=0 previously and here expand one?
```
fseasy, line 729: Here need author review. I don't know why should this, so I add a comment without change anything.

deepface/DeepFace.py
```diff
@@ -702,23 +716,23 @@ def represent(
...
     else:  # skip
-        if isinstance(img_path, str):
-            img = functions.load_image(img_path)
-        elif type(img_path).__module__ == np.__name__:
+        if type(img_path).__module__ == np.__name__:
             img = img_path.copy()
         else:
-            raise ValueError(f"unexpected type for img_path - {type(img_path)}")
+            # Try load. If load error, will raise exception internal
+            img, _ = functions.load_image(img_path)
```
fseasy, line 723: Because load_image in fact support many type, and have properly raise. so here I just use `try-except` logic  -- just call it. if failed, it will raise, and it should has almost same effects.

serengil:
In the line 706 and 721, there are bugs but these are a few line of codes. Please just change these lines, this PR may cause unpredictable problems.

deepface/DeepFace.py
```diff
@@ -702,23 +716,23 @@ def represent(
             align=align,
         )
     else:  # skip
-        if isinstance(img_path, str):
-            img = functions.load_image(img_path)
-        elif type(img_path).__module__ == np.__name__:
+        if type(img_path).__module__ == np.__name__:
```
serengil, line 719: I think we do not need to check type. load_image handles loading different type of inputs. So, we can get rid of the if and else block here.

deepface/DeepFace.py
serengil: face recognition models expect inputs like (1, 224, 224, 3). so, this is intended behaviour. please remove your comment.

deepface/commons/functions.py
```diff
@@ -94,6 +94,14 @@ def load_image(img):
     if type(img).__module__ == np.__name__:
         return img, None
 
+    try:
+        # Test whether img is a Python3's Path. If hit, tranform to str to let following logic work.
+        from pathlib import Path
```
serengil, line 99: This is already imported in line 3: https://github.com/serengil/deepface/blob/master/deepface/commons/functions.py#L3C21-L3C25

This causes linting error in github actions:

************* Module deepface.commons.functions
deepface/commons/functions.py:99:8: W0621: Redefining name 'Path' from outer scope (line 3) (redefined-outer-name)
deepface/commons/functions.py:99:8: W0404: Reimport 'Path' (imported line 3) (reimported)

serengil:
Added some comments.

Also, linting step is failed: https://github.com/serengil/deepface/actions/runs/7205630424/job/19636358763?pr=917

deepface/DeepFace.py
fseasy: but Why we do `img = img[0]` in 726

fseasy:
> Added some comments.
> 
> Also, linting step is failed: https://github.com/serengil/deepface/actions/runs/7205630424/job/19636358763?pr=917

Thanks for quickly response. I'll change it.

deepface/DeepFace.py
fseasy: Previously I find the diffrence between here and `load_image` for copy. But after carefully look, no copy is ok. I'll do it.

fseasy:
updated. please check if you has time.

deepface/DeepFace.py
serengil: it is expanding. if your input is (224, 224, 3) then this makes (1, 224, 224, 3). your command drops a dimension on the other hand.

serengil:

PR #922, opened 2023-12-17
Adding Type Hints for Enhanced Code Clarity and Maintenance

RazaProdigy:
Objective: 
This pull request introduces type hints to the existing codebase for deepface.py. The addition of type hints aims to enhance code clarity, facilitate maintenance, and improve the development experience for contributors and users of the repository.

Key Changes:

- Added type hints to function parameters and return types across various modules.
- Ensured compatibility with existing code by performing thorough checks and adhering to the current logic and structure.
- Enhanced readability and understandability of the code, especially in complex functions.

No Breaking Changes: These additions are non-intrusive and backward compatible. They do not alter the logic or functionality of the existing code.

Request for Review:
I request the maintainers to review these changes. Feedback and suggestions for further refinement are highly appreciated.

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -5,6 +5,7 @@ import warnings
 import time
 import pickle
 import logging
+from typing import Any, Dict, List, Tuple
 
 # 3rd party dependencies
 import numpy as np
@@ -45,7 +46,7 @@ if tf_version == 2:
 # -----------------------------------
 
 
-def build_model(model_name):
+def build_model(model_name: str) -> Any:
     """
     This function builds a deepface model
     Parameters:
@@ -91,15 +92,15 @@ def build_model(model_name):
 
 
 def verify(
-    img1_path,
-    img2_path,
-    model_name="VGG-Face",
-    detector_backend="opencv",
-    distance_metric="cosine",
-    enforce_detection=True,
-    align=True,
-    normalization="base",
-):
+    img1_path: str,
+    img2_path: str,
+    model_name: str = "VGG-Face",
+    detector_backend: str = "opencv",
+    distance_metric: str = "cosine",
+    enforce_detection: bool = True,
+    align: bool = True,
+    normalization: str = "base",
+) -> Dict[str, Any]:
     """
     This function verifies an image pair is same person or different persons. In the background,
     verification function represents facial images as vectors and then calculates the similarity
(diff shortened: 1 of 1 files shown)


deepface/DeepFace.py
```diff
@@ -91,15 +92,15 @@ def build_model(model_name):
...
-    img1_path,
-    img2_path,
-    model_name="VGG-Face",
-    detector_backend="opencv",
-    distance_metric="cosine",
-    enforce_detection=True,
-    align=True,
-    normalization="base",
-):
+    img1_path: str,
```
serengil, line 95: all image path variables (img_path, img1_path, img2_path) can be string or numpy array. So, these should accept numpy array as

```python
from typing import Union
import numpy as np

def verify(img1_path: Union[str, np.ndarray]):
   pass
```

deepface/DeepFace.py
```diff
@@ -45,7 +46,7 @@
 # -----------------------------------
 
 
-def build_model(model_name):
+def build_model(model_name: str) -> Any:
```
serengil, line 49: build_model returns one of these objects:

- keras.engine.functional.Functional
- deepface.basemodels.DlibResNet.DlibResNet
- deepface.basemodels.SFace.SFaceModel

So, this should be typed as

```python
from typing import Union
from keras.engine.functional import Functional
from deepface.basemodels import DlibResNet, SFace

def build_model(model_name: str) -> Union[Functional, DlibResNet.DlibResNet, SFace.SFaceModel]:
    pass
```

We should then get rid of Any import in this module.

serengil: Or you may consider to use Functional and any. Because except 2 models, this function returns Functional.

```python
Union[Functional, Any]
```

RazaProdigy:
I've updated the type hints in the image paths and build_model function as per your suggestions. The new commit is pushed to this PR. Could you please take another look when you have a chance?

Thank you!

deepface/DeepFace.py
```diff
@@ -231,13 +234,13 @@ def verify(
...
 
 def analyze(
-    img_path,
-    actions=("emotion", "age", "gender", "race"),
-    enforce_detection=True,
-    detector_backend="opencv",
-    align=True,
-    silent=False,
-):
+    img_path: Union[str, np.ndarray],,
```
serengil, line 237: Double comma at the end of the line breaks the code

RazaProdigy: Ah, My eyes missed that. 
Apologies! Fixed now

PR #941, opened 2023-12-30
yolo-link-fixed

sbmalik:
Please upload the model on your [deepface_models](https://github.com/serengil/deepface_models). The url is added in the code but it is assuming that the model is already at deepface_model.


serengil:
I am just uploading h5 files that I exported to that repo. Uploading this file may cause copyright issue. Instead, I prefer to download it from external repo. 

For instance, [this repo](https://github.com/akanametov/yolov8-face) publishes that weight file in the url: https://github.com/akanametov/yolov8-face/releases/download/v0.0.0/yolov8n-face.pt

PR #1011, opened 2024-02-07
add features

wilson1231234:
## Tickets

### What has been done

With this PR, ...
add features for kyc

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,0 +1,8 @@
+services:
+  app-1:
+    build:
+      context: .
+    ports:
+      - "5000:5000"
+    volumes:
+      - ./:/app-1
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,8 +1,0 @@
-services:
-  app-1:
-    build:
-      context: .
-    ports:
-      - "5000:5000"
-    volumes:
-      - ./:/app-1
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -527,3 +527,7 @@ def detectFace(
     if len(face_objs) > 0:
         extracted_face = face_objs[0]["face"]
     return extracted_face
+
+
+
+
--- a/deepface/api/src/modules/core/routes.py
+++ b/deepface/api/src/modules/core/routes.py
@@ -19,7 +19,7 @@ def represent():
     if input_args is None:
         return {"message": "empty input set passed"}
 
-    img_path = input_args.get("img") or input_args.get("img_path")
+    img_path = input_args.get("img1_path")
     if img_path is None:
         return {"message": "you must pass img_path input"}
 
@@ -48,8 +48,8 @@ def verify():
     if input_args is None:
         return {"message": "empty input set passed"}
 
-    img1_path = input_args.get("img1") or input_args.get("img1_path")
-    img2_path = input_args.get("img2") or input_args.get("img2_path")
+    img1_path = input_args.get("img1_path")
+    img2_path = input_args.get("img2_path")
 
     if img1_path is None:
         return {"message": "you must pass img1_path input"}
@@ -73,8 +73,9 @@ def verify():
         enforce_detection=enforce_detection,
     )
 
+    verification["verified"] = str(verification["verified"])
+    del verification["facial_areas"]
     logger.debug(verification)
-
     return verification
 
 
@@ -85,9 +86,9 @@ def analyze():
     if input_args is None:
         return {"message": "empty input set passed"}
 
-    img_path = input_args.get("img") or input_args.get("img_path")
+    img_path = input_args.get("img1_path")
     if img_path is None:
-        return {"message": "you must pass img_path input"}
+        return {"message": "you must pass img1_path input"}
 
     detector_backend = input_args.get("detector_backend", "opencv")
     enforce_detection = input_args.get("enforce_detection", True)
@@ -102,6 +103,10 @@ def analyze():
         align=align,
     )
 
+    del demographies["results"][0]["region"]
+    del demographies["results"][0]["emotion"]
+    del demographies["results"][0]["race"]
+    del demographies["results"][0]["gender"]
+    del demographies["results"][0]["face_confidence"]
     logger.debug(demographies)
-
-    return demographies
+    return demographies["results"][0]
--- a/deepface/api/src/modules/core/routes.py
+++ b/deepface/api/src/modules/core/routes.py
@@ -59,7 +59,6 @@ def verify():
 
     model_name = input_args.get("model_name", "VGG-Face")
     detector_backend = input_args.get("detector_backend", "opencv")
-    print("using detector_backend: " + detector_backend)
     enforce_detection = input_args.get("enforce_detection", True)
     distance_metric = input_args.get("distance_metric", "cosine")
     align = input_args.get("align", True)
@@ -94,7 +93,6 @@ def analyze():
         return {"message": "you must pass img_path input"}
 
     detector_backend = input_args.get("detector_backend", "opencv")
-    print("using detector_backend: " + detector_backend)
     enforce_detection = input_args.get("enforce_detection", True)
     align = input_args.get("align", True)
     actions = input_args.get("actions", ["age", "gender", "emotion", "race"])
--- a/deepface/api/src/modules/core/routes.py
+++ b/deepface/api/src/modules/core/routes.py
@@ -19,7 +19,7 @@ def represent():
     if input_args is None:
         return {"message": "empty input set passed"}
 
-    img_path = input_args.get("img1_path")
+    img_path = input_args.get("img") or input_args.get("img_path")
     if img_path is None:
         return {"message": "you must pass img_path input"}
 
@@ -48,8 +48,8 @@ def verify():
     if input_args is None:
         return {"message": "empty input set passed"}
 
-    img1_path = input_args.get("img1_path")
-    img2_path = input_args.get("img2_path")
+    img1_path = input_args.get("img1") or input_args.get("img1_path")
+    img2_path = input_args.get("img2") or input_args.get("img2_path")
 
     if img1_path is None:
         return {"message": "you must pass img1_path input"}
@@ -73,9 +73,8 @@ def verify():
         enforce_detection=enforce_detection,
     )
 
-    verification["verified"] = str(verification["verified"])
-    del verification["facial_areas"]
     logger.debug(verification)
+
     return verification
 
 
@@ -86,9 +85,9 @@ def analyze():
     if input_args is None:
         return {"message": "empty input set passed"}
 
-    img_path = input_args.get("img1_path")
+    img_path = input_args.get("img") or input_args.get("img_path")
     if img_path is None:
-        return {"message": "you must pass img1_path input"}
+        return {"message": "you must pass img_path input"}
 
     detector_backend = input_args.get("detector_backend", "opencv")
     enforce_detection = input_args.get("enforce_detection", True)
@@ -103,10 +102,6 @@ def analyze():
         align=align,
     )
 
-    del demographies["results"][0]["region"]
-    del demographies["results"][0]["emotion"]
-    del demographies["results"][0]["race"]
-    del demographies["results"][0]["gender"]
-    del demographies["results"][0]["face_confidence"]
     logger.debug(demographies)
-    return demographies["results"][0]
+
+    return demographies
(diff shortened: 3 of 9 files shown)


compose.yaml
```diff
@@ -0,0 +1,8 @@
+services:
```
serengil, line 1: we don't need compose in this project, please remove this

one can do anything with docker run

deepface/modules/preprocessing.py
```diff
@@ -72,6 +78,14 @@ def load_base64(uri: str) -> np.ndarray:
     # img_rgb = cv2.cvtColor(img_bgr, cv2.COLOR_BGR2RGB)
     return img_bgr
 
+def loadVideo_base64(uri: str) -> np.ndarray:
```
serengil, line 81: function name should follow snake case instead of camel case

deepface/modules/preprocessing.py
```diff
@@ -129,3 +143,58 @@ def normalize_input(img: np.ndarray, normalization: str = "base") -> np.ndarray:
         raise ValueError(f"unimplemented normalization type - {normalization}")
 
     return img
+
+
+def extract_frames_with_faces(video_bytes):
```
serengil, line 148: type of video_bytes and return type of this function are missing

deepface/modules/preprocessing.py
```diff
@@ -33,6 +35,10 @@ def load_image(img: Union[str, np.ndarray]) -> Tuple[np.ndarray, str]:
     # The image is a base64 string
     if img.startswith("data:image/"):
         return load_base64(img), "base64 encoded string"
+    
+    # The video is a base64 string
+    if img.startswith("data:video/"):
+        return loadVideo_base64(img), "base64 encoded string"
```
serengil, line 41: load_image's return type is numpy array or string, but load video base 64 is returning list of string.

deepface/modules/preprocessing.py
```diff
@@ -129,3 +143,58 @@ def normalize_input(img: np.ndarray, normalization: str = "base") -> np.ndarray:
...
+    video_bytes: The video bytes.
+
+  Returns:
+    A list of base64-encoded frames with faces.
+  """
+  video_bytes = base64.b64decode(video_bytes)
+
+  # Create a file-like object to write the video to.
+  hash = str(random.getrandbits(128))
+  with io.open("./"+ hash + ".mp4", "wb") as f:
```
serengil, line 161: you may consider to set a variable like target_file and use same variable while deleting the file

deepface/modules/preprocessing.py
```diff
@@ -129,3 +143,58 @@ def normalize_input(img: np.ndarray, normalization: str = "base") -> np.ndarray:
...
+    f.write(video_bytes)
+  # Decode the video bytes.
+  video_capture = cv2.VideoCapture("./"+ hash + ".mp4")
+
+  # Create a face detector.
+  face_detector = cv2.CascadeClassifier(cv2.data.haarcascades + 'haarcascade_frontalface_default.xml')
```
serengil, line 167: DeepFace.extract_faces should be used here instead of custom opencv implementation

deepface/modules/preprocessing.py
```diff
@@ -129,3 +143,58 @@ def normalize_input(img: np.ndarray, normalization: str = "base") -> np.ndarray:
...
+    # Detect faces in the frame.
+    faces = face_detector.detectMultiScale(gray, 1.3, 5)
+
+    # If there are faces in the frame, add it to the list of frames with faces.
+    if len(faces) > 0:
+      # Encode the frame as a base64 string.
+      frame_bytes = cv2.imencode('.jpg', frame)[1].tobytes()
+      frame_base64 = base64.b64encode(frame_bytes).decode('utf-8')
+      frames_with_faces.append(frame_base64)
+      break
```
serengil, line 194: what if source frame has more than one image? not good practice to work with the first face.

deepface/modules/preprocessing.py
serengil, line 41: return type of load image is image itself as numpy array. load video base64 returns that numpy array as many numpy arrays.

Won't this fail here: https://github.com/serengil/deepface/blob/master/deepface/modules/detection.py#L77C5-L77C8

deepface/modules/preprocessing.py
```diff
@@ -129,3 +143,58 @@ def normalize_input(img: np.ndarray, normalization: str = "base") -> np.ndarray:
...
+        frames_with_faces.append(frame_base64)
+        break
+    else:
+        tempFrame = frame
+
+    # Convert the frame to grayscale.
+    gray = cv2.cvtColor(frame, cv2.COLOR_BGR2GRAY)
+
+    # Detect faces in the frame.
+    faces = face_detector.detectMultiScale(gray, 1.3, 5)
```
serengil, line 186: once load image is called, face detector module is being called already. why we need to call face extraction here?

https://github.com/serengil/deepface/blob/master/deepface/modules/detection.py#L84

wilson1231234:
Thanks for your comments, this branch is for loading the selfie video, and then extract the face frame of user, verify the face with user's face on ID card. For now, first face is enough, but may have to extract more faces in video to increase the accuracy. Thus, the cases of more than 1 person's face in video will not be consider.

serengil:
I tested the behaviour after load_image: https://github.com/serengil/deepface/blob/master/deepface/modules/detection.py#L84C21-L84C36

DetectorWrapper.detect_faces is not working many numpy array inputs for many faces.

serengil:
This cannot be merged to the source code.

deepface/modules/preprocessing.py
wilson1231234: Thanks for reminding, will be update in the next version.

deepface/modules/preprocessing.py
wilson1231234: cv2.VideoCapture cannot load the bytes directly, I tried using io to pretend the video bytes like streaming video but failed. Thus, I have to store at disk as temp file to load it. Do you have better suggestion? thanks

serengil: i mean 

target_file = "./"+ hash + ".mp4"
with io.open(target_file , "wb") as f:
...
os.remove(target_file)

wilson1231234: sure, will be update in the next version, thanks

wilson1231234:
add features for kyc purpose, 
{
   "img1_path":"identity_data:image/base64,
   "img2_path":"selfie_data:image/base64,
}

for selfie_data, more than 1 face in same frame will throw error, identity_data card will not.

deepface/DeepFace.py
```diff
@@ -527,3 +527,7 @@ def detectFace(
     if len(face_objs) > 0:
         extracted_face = face_objs[0]["face"]
     return extracted_face
+
```
serengil, line 530: no changes in this file, please revert

deepface/api/src/modules/core/routes.py
```diff
@@ -78,6 +78,9 @@ def verify():
     return verification
 
 
+
```
serengil, line 81: no changes in this file, please revert

deepface/api/src/modules/core/service.py
```diff
@@ -52,3 +52,5 @@ def analyze(img_path, actions, detector_backend, enforce_detection, align):
         return result
     except Exception as err:
         return {"error": f"Exception while analyzing: {str(err)}"}, 400
+
```
serengil, line 55: no changes in this file, please revert

deepface/modules/demography.py
```diff
@@ -126,6 +127,16 @@ def analyze(
         expand_percentage=expand_percentage,
     )
 
+    #kyc checking
+    if isinstance(img_path, str):
```
serengil, line 131: this should be checked in extract_faces because of separation of concern

deepface/modules/verification.py
```diff
@@ -105,6 +105,22 @@ def verify(
         align=align,
         expand_percentage=expand_percentage,
     )
+
+    #kyc checking
+    if isinstance(img1_path, str):
```
serengil, line 110: this should be checked in extract_faces because of separation of concern

deepface/modules/verification.py
```diff
@@ -105,6 +105,22 @@ def verify(
...
+        if img1_path.startswith("selfie_data:image/"):
+            #If more than 1 face in frame, violate kyc rules, throw exception
+            if len(img1_objs) > 1:
+                raise RuntimeError('More than 1 person detected: len(img1_objs) > 1')
+            
+    #kyc checking
+    if isinstance(img2_path, str):
```
serengil, line 117: this should be checked in extract_faces because of separation of concern

deepface/modules/preprocessing.py
```diff
@@ -34,6 +36,19 @@ def load_image(img: Union[str, np.ndarray]) -> Tuple[np.ndarray, str]:
     if img.startswith("data:image/"):
         return load_base64(img), "base64 encoded string"
 
+    # The kyc image is a base64 string
+    if img.startswith("selfie_data:image/"):
```
serengil, line 40: this can be added in line 36's if statatement as an or condition

deepface/modules/preprocessing.py
```diff
@@ -34,6 +36,19 @@ def load_image(img: Union[str, np.ndarray]) -> Tuple[np.ndarray, str]:
...
+        return load_base64(img), "base64 encoded string"   
+    
+    # The kyc image is a base64 string
```
serengil, line 43: this can be added in line 36's if statatement as an or condition

deepface/modules/preprocessing.py
```diff
@@ -2,12 +2,14 @@
 from typing import Union, Tuple
 import base64
 from pathlib import Path
-
 # 3rd party
 import numpy as np
 import cv2
 from PIL import Image
 import requests
+from deepface.modules.kyc_validation import extract_frame_with_face
```
serengil, line 10: what does kyc mean?

deepface/modules/kyc_validation.py
```diff
@@ -0,0 +1,148 @@
...
+import random
+from typing import List
+import asyncio
+
+
+import cv2
+from deepface import DeepFace
+import numpy as np
+
+detector_backend_extract = "retinaface"
```
serengil, line 13: detector names are being set in extract_faces, we should use the one set instead of static one

deepface/modules/kyc_validation.py
```diff
@@ -0,0 +1,148 @@
...
+      elif len(faces) > 1:
+        raise RuntimeError('More than 1 person detected: len(faces) > 1')
+
+       
+
+    frame_counter = frame_counter + 1
+
+  # Close the video capture.
+  video_capture.release()
+  """ Accuracy problem
```
serengil, line 90: uncommented block should not be available in the code

deepface/modules/kyc_validation.py
```diff
@@ -0,0 +1,148 @@
...
+  if len(frames_with_faces) > 1:
+    result = asyncio.run(check_same_face(frames_with_faces))
+    if result == False:
+       raise RuntimeError('More than 1 person detected: check_same_face')
+  """
+
+  return frames_with_faces[0]
+
+
+async def check_same_face(frames_with_faces: List[str]) -> bool:
```
serengil, line 101: this is not being used, please get rid of it

deepface/modules/kyc_validation.py
```diff
@@ -0,0 +1,148 @@
...
+    pair_list = generate_pairs(frames_with_faces)
+
+    tasks = [asyncio.create_task(call_verify_async(pair)) for pair in pair_list]
+
+    results = await asyncio.gather(*tasks)
+    print(str(results))
+    return all(results)
+
+
+async def call_verify_async(pairs:List[str]) -> bool:
```
serengil, line 120: this is not being used, please get rid of it

deepface/modules/kyc_validation.py
```diff
@@ -0,0 +1,148 @@
...
+
+    obj = DeepFace.verify(
+    img1_path=pairs[0],
+    img2_path=pairs[1],
+    detector_backend = detector_backend_verify
+    )
+    return obj["verified"]
+
+
+def generate_pairs(imgs:List[str]) -> List[List[str]]:
```
serengil, line 130: this is not being used, please get rid of it

serengil:
I added lots of comments.

Still, not sure this is going to be helpful feature for deepface. Maybe instead of merging this, you should use this feature in your own fork.

serengil:
IMO, you should have this feature in your own branch

deepface/modules/preprocessing.py
wilson1231234: Know Your Client. KYC or KYC check is the mandatory process of identifying and verifying the client's identity when opening an account and periodically over time. In other words, banks must ensure that their clients are genuinely who they claim to be.

wilson1231234:
It is fine, thanks for your suggestions and commends.

Issue #1021, opened 2024-02-12
Align first and detect second in detection module by default


serengil:
Currently, we are detecting facial area first and align it second. This causes to have many black pixels in the input image.

Similar to RetinaFace, we can align first and detect second to get rid of those black pixels - if there is one face in the image.

## Recommendation

Maybe detector implementations should return facial area coordinates and eye coordinates as output, and we should do detection & alignment in the parent module. Otherwise, we need to copy same copy to all detectors.

Similarly, expand facial area will be done in the parent module. Detectors are just responsible for findind x, y, w, h and eye locations.

Concern: dlib has its own alignment module and it is different than others. We may find eye locations with dlib and apply alignment similar to other detectors.

```python
shape_predictor_path = 'path/to/shape_predictor_5_face_landmarks.dat'
predictor = dlib.shape_predictor(shape_predictor_path)
shape = predictor(gray, face)
# Extract coordinates of the left and right eyes
left_eye = (shape.part(2).x, shape.part(2).y)  # Change index if needed
right_eye = (shape.part(0).x, shape.part(0).y)  # Change index if needed
```

## To-Dos

- [x] change detector interface, instead detect_faces, add detect_facial_areas and return and object with x, y, h, w, left_eye, right_eye and confidence score
- [x] do alignment in the parent module
- [x] do expand facial area in the parent module
- [x] do alignment first and detection second if there is just one face
- [x] resize should be skipped if target_size is set to None - None should be accepted in type hinting, too.

Refs:
- https://github.com/serengil/retinaface/blob/master/retinaface/RetinaFace.py#L282
- https://github.com/serengil/retinaface/blob/master/retinaface/commons/postprocess.py#L96

serengil:
Closed with PR - https://github.com/serengil/deepface/pull/1025

serengil:
Comparative results are shown in the following illustration

![align-first](https://github.com/serengil/deepface/assets/18491038/18a96946-8ebd-4896-a371-e564867ed2b2)

PR #1025, opened 2024-02-16
Feat task 1602 new detector interface

serengil:
## Tickets

https://github.com/serengil/deepface/issues/1021

### What has been done

With this PR,

- Detector interface is changed. Now, all detectors are returning facial area coordinates, eye coordinates and confidence score. They are not responsible for extracting face, alignment or expanding facial area.
- Extracting face, alignment and expanding facial area are now done in DetectorWrapper which is the parent object of those detectors.
- Besides, we are now align first and extract facial area second not to have many black pixels.

## How to test

```shell
make lint && make test
```


serengil:
Comparative results are shown in the following illustration

align-first

PR #1031, opened 2024-02-21
Small knob polishing and amend duplicated attempts to add undetectable images to pickle

### LINKED ISSUE #1030: Find match from dataset continuosly try to add unrecognized images to pickle
Just cloned and set up the project locally (Windows 11)
Run `python  .\stream_py` from tests directory and pickle file `representations_vgg_face.pkl` get built.

image

For whatever reason the image file img39.jpg doesn't get any representation hence not added to the pickle file.
As a result the following log line keeps popping

```
24-02-21 18:28:25 - ⚠️ Items ['dataset\\img39.jpg'] were added into dataset just after data source dataset\representations_vgg_face.pkl created!
```

Which is actually not true.
IMO for every processed file would be worth adding a reference into the pickle file with null representation and all values to zero.

AndreaLanfranchi:
Hope you don't mind some small adjustments
Fixes #1030 

Comments welcome.

### DIFF
--- a/deepface/modules/recognition.py
+++ b/deepface/modules/recognition.py
@@ -391,11 +391,11 @@ def __find_bulk_embeddings(
 
                 img_representation = embedding_obj[0]["embedding"]
                 representations.append((
-                    employee, 
-                    img_representation, 
-                    img_region["x"], 
-                    img_region["y"], 
-                    img_region["w"], 
+                    employee,
+                    img_representation,
+                    img_region["x"],
+                    img_region["y"],
+                    img_region["w"],
                     img_region["h"]
                     ))
 
--- a/deepface/modules/recognition.py
+++ b/deepface/modules/recognition.py
@@ -99,7 +99,7 @@ def find(
 
     file_name = f"representations_{model_name}.pkl"
     file_name = file_name.replace("-", "_").lower()
-    datastore_path = f"{db_path}/{file_name}"
+    datastore_path = os.path.join(db_path, file_name)
 
     df_cols = [
         "identity",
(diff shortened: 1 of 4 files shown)


deepface/modules/recognition.py
```diff
@@ -292,7 +295,7 @@ def find(
     return resp_obj
 
 
-def __list_images(path: str) -> list:
+def __list_images(path: str) -> list[str]:
```
serengil, line 298: This line causes failing of unit tests: https://github.com/serengil/deepface/actions/runs/7993768781/job/21857992920?pr=1031

This should be changed to List[str] if you insist to use string here.

AndreaLanfranchi: Thank you. Done.

deepface/modules/recognition.py
```diff
@@ -304,11 +306,11 @@ def __list_images(path: str) -> list[str]:
         images (list): list of exact image paths
     """
     images = []
-    for r, _, f in os.walk(path):
-        for file in f:
-            if file.lower().endswith((".jpg", ".jpeg", ".png")):
-                exact_path = os.path.join(r, file)
-                images.append(exact_path)
+    pattern = re.compile(r".*\.(jpg|jpeg|png)$", re.IGNORECASE)
```
serengil, line 309: prefer not to use regex, would you please restore this block to its old version.

AndreaLanfranchi: Done ... just fyi the regex version shaves about 20milliseconds from every loop.

serengil:
LGTM

PR #1032, opened 2024-02-22
Simplify find data initialization

AndreaLanfranchi:
This simplifies the cyclomatic complexit of data initialization in `find` function.

- Ensures a pickle file always exists (although might be empty)
- Removes two `if` branches and duplicated code
- Early bailout if no representations

Thank you for your patience and attention.

### DIFF
--- a/deepface/modules/recognition.py
+++ b/deepface/modules/recognition.py
@@ -100,6 +100,7 @@ def find(
     file_name = f"representations_{model_name}.pkl"
     file_name = file_name.replace("-", "_").lower()
     datastore_path = os.path.join(db_path, file_name)
+    representations = []
 
     df_cols = [
         "identity",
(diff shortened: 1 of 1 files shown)


deepface/modules/recognition.py
```diff
@@ -110,93 +111,71 @@ def find(
...
-        alpha_employees = __list_images(path=db_path)
-        beta_employees = [representation[0] for representation in representations]
-
-        newbies = list(set(alpha_employees) - set(beta_employees))
-        oldies = list(set(beta_employees) - set(alpha_employees))
+    # Ensure the proper pickle file exists
+    if not os.path.exists(datastore_path):
+        with open(datastore_path, "wb") as f:
+            pickle.dump([], f)
+            f.close()
```
serengil, line 118: you do not have to use close if you use with statement.

deepface/modules/recognition.py
```diff
@@ -110,93 +111,71 @@ def find(
...
-                enforce_detection=enforce_detection,
-                align=align,
-                normalization=normalization,
-                silent=silent,
-            )
-            representations = representations + newbies_representations
+    # Load the representations from the pickle file
+    with open(datastore_path, "rb") as f:
+        representations = pickle.load(f)
+        f.close()
```
serengil, line 123: you do not have to use close if you use with statement.

deepface/modules/recognition.py
AndreaLanfranchi: Thank you. Done.

(still learning semantics)

serengil:
LGTM

PR #1034, opened 2024-02-23
refactor: simplify the inception resnet v2 model definition code

bpradana:
### What has been done

With this PR, I have simplify the code for InceptionResNetV2 model. I remove the repetition in the code and replace it with for loop for better readbility

## How to test

I have compare it with the original code and it performs exactly the same


serengil:
I converted from loop approach to that one already. Because I found this is more legible. Thank you but I prefer not to merge this.

bpradana:
thank you for the feedback!

AndreaLanfranchi:
Hey @bpradana I am interested into your pr.
Would open a PR in my repo ?

bpradana:
@AndreaLanfranchi sure!

PR #1039, opened 2024-02-24
Handle ValueError exception in find

AndreaLanfranchi:
When exctract_faces is invoked with  enforce_detection == True is useful to treat the exception as a not found.

### DIFF
--- a/deepface/modules/recognition.py
+++ b/deepface/modules/recognition.py
@@ -181,16 +181,19 @@ def find(
         columns=df_cols,
     )
 
-    # img path might have more than once face
-    source_objs = detection.extract_faces(
-        img_path=img_path,
-        target_size=target_size,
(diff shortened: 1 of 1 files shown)


serengil:
When I call find function, I want to see the error mentioning no face detected instead of empty list. You should do this exception handling where find function is called.

AndreaLanfranchi:
Gotcha.
However, for my knowledge, what is the point of throwing an exception on a function which finds ? An empty results set is more than enough (unless other errors occur). Eventually the function can be used as the inverse i.e. asserting there are no faces or no matches. It would be up the callee to decide how to treat the return result (not the caller to decide).
Just for the sake of knowledge...

serengil:
Because empty list means that no matching face found. On the other hand, it can still find matching faces if you set enforce detection to false.

PR #1040, opened 2024-02-25
Correct info about returned types for the extract_faces method

DriesVerachtert:
### What has been done

The 2 extract_faces methods have documentation which explain the types that are returned. This documentation doesn't seem to be correct (anymore). This PR just updates the method documentation.

## How to test

No actual code has been changed.


serengil:
That is my bad 😅 Thank you for your contribution 🙏

PR #1044, opened 2024-02-26
Correct info about returned types for the represent method

DriesVerachtert:
### What has been done

Small fix to the documentation of the represent method: the embedding is returned as a list of floats, not a multidimensional array.

### How to test

The code hasn't been changed, only the documentation

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -345,7 +345,7 @@ def represent(
         results (List[Dict[str, Any]]): A list of dictionaries, each containing the
             following fields:
 
-        - embedding (np.array): Multidimensional vector representing facial features.
+        - embedding (List[float]): Multidimensional vector representing facial features.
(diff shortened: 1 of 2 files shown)


serengil:
embeddings are not single dimensional vectors, e.g. facenet's output is 128 in other words it is 128 dimensional vector

DriesVerachtert:
Ok I'll change that part back to 'multi-dimensional' in the MR.

Maybe the meaning of multi-dimensional is slightly different between languages. In Dutch we would describe a single row of values or a single list of a bunch of values as being one-dimensional. If there would be multiple rows and multiple columns, then we would call in two-dimensional.

serengil:
This is a concept more from vectors in physics rather than language.

https://en.m.wikipedia.org/wiki/Vector_(mathematics_and_physics)

PR #1047, opened 2024-02-27
Simplify and strengthen rotate_facial_area

### LINKED ISSUE #1045: Expand facial area logic is expanding the facial area twice of expected
As @AndreaLanfranchi raised, current expand facial area logic is expanding w and h 2 * expand_percentage instead of expand_percentage.

Also, one can understand that w * h should be expanded with the given value, but we are aiming to expand w and h separately with that given value. This should be explained in the docstring.

Ref: https://github.com/serengil/deepface/blob/master/deepface/detectors/DetectorWrapper.py#L106

AndreaLanfranchi:
This strengthen `rotate_facial_area` by :
 - constraining rotation angle in range [0, 360)
 - early bailout in case angle == 0
 - deriving direction of rotation from sign of  `angle` : positive values for clockwise, negative for anti-clockwise

In previous implementation direction could be any number causing unwanted results.

Also fixes #1045

### DIFF
--- a/deepface/detectors/DetectorWrapper.py
+++ b/deepface/detectors/DetectorWrapper.py
@@ -124,11 +124,13 @@ def detect_faces(
                 img=img, left_eye=left_eye, right_eye=right_eye
             )
             rotated_x1, rotated_y1, rotated_x2, rotated_y2 = rotate_facial_area(
-                facial_area=(x, y, x + w, y + h), 
-                angle=angle, 
+                facial_area=(x, y, x + w, y + h),
+                angle=angle,
                 size=(img.shape[0], img.shape[1])
             )
-            detected_face = aligned_img[int(rotated_y1) : int(rotated_y2), int(rotated_x1) : int(rotated_x2)]
+            detected_face = aligned_img[
+                int(rotated_y1) : int(rotated_y2),
+                int(rotated_x1) : int(rotated_x2)]
 
         result = DetectedFace(
             img=detected_face,
@@ -142,7 +144,7 @@ def detect_faces(
 
 
 def rotate_facial_area(
-    facial_area: Tuple[int, int, int, int], 
+    facial_area: Tuple[int, int, int, int],
     angle: float,               # in degrees. The sign determines the direction of rotation
     size: Tuple[int, int]       # (width, height)
 ) -> Tuple[int, int, int, int]:
--- a/deepface/detectors/DetectorWrapper.py
+++ b/deepface/detectors/DetectorWrapper.py
@@ -104,18 +104,6 @@ def detect_faces(
         confidence = facial_area.confidence
 
         if expand_percentage > 0:
-            # Uncomment this if you want to :
-            # Expand the facial area to be extracted and recompute the height and width
-            # keeping the same aspect ratio and ensuring that the expanded area stays
-            # within img.shape limits
-
-            # current_area = w * h
-            # expanded_area = current_area + int((current_area * expand_percentage) / 100)
-            # scale_factor = math.sqrt(expanded_area / current_area)
-            # expanded_w = int(w * scale_factor)
-            # expanded_h = int(h * scale_factor)
-
-            # Or uncomment this if you want to :
             # Expand the facial region height and width by the provided percentage
             # ensuring that the expanded region stays within img.shape limits
             expanded_w = w + int(w * expand_percentage / 100)
--- a/deepface/detectors/DetectorWrapper.py
+++ b/deepface/detectors/DetectorWrapper.py
@@ -129,12 +129,6 @@ def detect_faces(
         # extract detected face unaligned
         detected_face = img[int(y) : int(y + h), int(x) : int(x + w)]
 
-        # aligning detected face causes a lot of black pixels
-        # if align is True:
-        #     detected_face, _ = detection.align_face(
-        #         img=detected_face, left_eye=left_eye, right_eye=right_eye
-        #     )
-
         # align original image, then find projection of detected face area after alignment
         if align is True:  # and left_eye is not None and right_eye is not None:
             aligned_img, angle = detection.align_face(
--- a/deepface/detectors/DetectorWrapper.py
+++ b/deepface/detectors/DetectorWrapper.py
@@ -123,10 +123,12 @@ def detect_faces(
             aligned_img, angle = detection.align_face(
                 img=img, left_eye=left_eye, right_eye=right_eye
             )
-            x1_new, y1_new, x2_new, y2_new = rotate_facial_area(
-                facial_area=(x2, y2, x2 + w2, y2 + h2), angle=angle, direction=1, size=img.shape
+            rotated_x1, rotated_y1, rotated_x2, rotated_y2 = rotate_facial_area(
+                facial_area=(x, y, x + w, y + h), 
+                angle=angle, 
+                size=(img.shape[0], img.shape[1])
             )
-            detected_face = aligned_img[int(y1_new) : int(y2_new), int(x1_new) : int(x2_new)]
+            detected_face = aligned_img[int(rotated_y1) : int(rotated_y2), int(rotated_x1) : int(rotated_x2)]
 
         result = DetectedFace(
             img=detected_face,
@@ -140,7 +142,9 @@ def detect_faces(
 
 
 def rotate_facial_area(
-    facial_area: Tuple[int, int, int, int], angle: float, direction: int, size: Tuple[int, int]
+    facial_area: Tuple[int, int, int, int], 
+    angle: float,               # in degrees. The sign determines the direction of rotation
+    size: Tuple[int, int]       # (width, height)
 ) -> Tuple[int, int, int, int]:
     """
     Rotate the facial area around its center.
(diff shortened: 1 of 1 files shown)


deepface/detectors/DetectorWrapper.py
```diff
@@ -123,10 +123,14 @@ def detect_faces(
             aligned_img, angle = detection.align_face(
                 img=img, left_eye=left_eye, right_eye=right_eye
             )
-            x1_new, y1_new, x2_new, y2_new = rotate_facial_area(
-                facial_area=(x2, y2, x2 + w2, y2 + h2), angle=angle, direction=1, size=img.shape
+            rotated_x1, rotated_y1, rotated_x2, rotated_y2 = rotate_facial_area(
+                facial_area=(x, y, x + w, y + h),
```
serengil, line 127: we want to rotate the expanded facial area. instead x, you should pass x2 similar to previous implementation.

AndreaLanfranchi: Oh ok. You're right.
Forgot I amended the caller too. 
Doing it now

deepface/detectors/DetectorWrapper.py
```diff
@@ -103,30 +103,45 @@ def detect_faces(
...
+            # expanded_w = int(w * scale_factor)
+            # expanded_h = int(h * scale_factor)
+
+            # Or uncomment this if you want to :
+            # Expand the facial region height and width by the provided percentage
+            # ensuring that the expanded region stays within img.shape limits
+            expanded_w = int(w * expand_percentage / 100)
+            expanded_h = int(h * expand_percentage / 100)
+
+            x = max(0, x - int((expanded_w - w) / 2))
```
serengil, line 124: this should be `x = max(0, x - int(expanded_w / 2))`

deepface/detectors/DetectorWrapper.py
```diff
@@ -103,30 +103,45 @@ def detect_faces(
...
+            y = max(0, y - int((expanded_h - h) / 2))
```
serengil, line 125: this should be ` y = max(0, y - int((expanded_h) / 2))`

deepface/detectors/DetectorWrapper.py
```diff
@@ -103,30 +103,45 @@ def detect_faces(
...
+            w = min(img.shape[1] - x, expanded_w)
+            h = min(img.shape[0] - y, expanded_h)
```
serengil, line 127: this should be `h = min(img.shape[0], h + 2 * expanded_h)`

deepface/detectors/DetectorWrapper.py
```diff
@@ -103,30 +103,45 @@ def detect_faces(
         right_eye = facial_area.right_eye
         confidence = facial_area.confidence
 
-        # expand the facial area to be extracted and stay within img.shape limits
-        x2 = max(0, x - int((w * expand_percentage) / 100))  # expand left
-        y2 = max(0, y - int((h * expand_percentage) / 100))  # expand top
-        w2 = min(img.shape[1], w + int((w * 2 * expand_percentage) / 100))  # expand right
-        h2 = min(img.shape[0], h + int((h * 2 * expand_percentage) / 100))  # expand bottom
+        if expand_percentage > 0:
+            # Uncomment this if you want to :
```
serengil, line 107: you can get rid of this unused block

deepface/detectors/DetectorWrapper.py
```diff
@@ -140,7 +155,9 @@ def detect_faces(
 
 
 def rotate_facial_area(
-    facial_area: Tuple[int, int, int, int], angle: float, direction: int, size: Tuple[int, int]
+    facial_area: Tuple[int, int, int, int],
+    angle: float,               # in degrees. The sign determines the direction of rotation
```
serengil, line 159: we should not have comment in this line, this is already explained in its docstring

deepface/detectors/DetectorWrapper.py
```diff
@@ -140,7 +155,9 @@ def detect_faces(
...
+    size: Tuple[int, int]       # (width, height)
```
serengil, line 160: similarly, we should get rid of the comment

deepface/detectors/DetectorWrapper.py
```diff
@@ -149,14 +166,22 @@ def rotate_facial_area(
...
         size (tuple of int): Tuple representing the size of the image (width, height).
 
     Returns:
         rotated_coordinates (tuple of int): Representing the new coordinates
             (x1, y1, x2, y2) or (x1, y1, x1+w1, y1+h1) of the rotated facial area.
     """
+
+    # Normalize the witdh of the angle so we don't have to
+    # worry about rotations greater than 360 degrees
+    angle = angle % 360
```
serengil, line 180: angle will be positive always because of the module, so you can get rid of the direction because it will be 1 always

deepface/detectors/DetectorWrapper.py
AndreaLanfranchi: Damn you're right.
I kept the code above from area expansion where the scale factor gave me the new w and h
Correcting.

deepface/detectors/DetectorWrapper.py
AndreaLanfranchi: That's not right.
75 % 360 == 75
-75 % 360 == -75

serengil: This is my experiment.

```shell
Python 3.9.16 (main, Jan 11 2023, 16:05:54)
[GCC 11.2.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = -75
>>> a % 360
285
>>>
>>>
>>> b = 75
>>> b % 360
```

serengil: still, both outputs will work as is

never mind this

AndreaLanfranchi: Ouch ... python mod operator is pretty fucked up. Every other language and calculators keeps the sign of dividend not the divisor.

AndreaLanfranchi: image

serengil:
Would you please run this code and share its output here

```python
import matplotlib.pyplot as plt

img_paths = ["dataset/dataset/img11.jpg", "dataset/img11_reflection.jpg"]

for img_path in img_paths:
	results = DeepFace.extract_faces(img_path = img_path)
	img = results[0]["face"]
	plt.imshow(img)
	plt.show()
```

deepface/detectors/DetectorWrapper.py
```diff
@@ -149,14 +154,22 @@ def rotate_facial_area(
...
+    if angle == 0:
+        return facial_area # No rotation needed
+    direction = 1 if angle > 0 else -1
```
serengil, line 171: direction must be equal to 1 always because if angle is negative then rotate sign will disappear if we set this to negative.

I cannot produce the problem because angle % 360 returns positive in my environment, but if i comment it for angle = -16, then dataset/img11_reflection.jpg is not working. this is sorted when i set direction to 1 by default.

serengil:
When I perform current implementation for dataset/img11_reflection.jpg, it is output is shown below - this is not correct!

This is sorted when i set direction to 1 by default

AndreaLanfranchi:
Just pushed a change which should correct the error.
This is the output you requested

image

serengil:
It is the output for img11.jpg, I wonder the output of img11_reflection.jpg also.

AndreaLanfranchi:
Yep ... sorry

image

deepface/detectors/DetectorWrapper.py
```diff
@@ -149,14 +154,24 @@ def rotate_facial_area(
...
             (x1, y1, x2, y2) or (x1, y1, x1+w1, y1+h1) of the rotated facial area.
     """
+
+    # Normalize the witdh of the angle so we don't have to
+    # worry about rotations greater than 360 degrees.
+    # We workaround the quirky behavior of the modulo operator
+    # for negative angle values.
+    direction = (1, -1)[angle < 0]
+    angle = abs(angle) % 360
+    if angle == 0:
```
serengil, line 172: should not we check this as a first task in this function?

deepface/detectors/DetectorWrapper.py
```diff
@@ -149,14 +154,24 @@ def rotate_facial_area(
...
     Returns:
         rotated_coordinates (tuple of int): Representing the new coordinates
             (x1, y1, x2, y2) or (x1, y1, x1+w1, y1+h1) of the rotated facial area.
     """
+
+    # Normalize the witdh of the angle so we don't have to
+    # worry about rotations greater than 360 degrees.
+    # We workaround the quirky behavior of the modulo operator
+    # for negative angle values.
+    direction = (1, -1)[angle < 0]
```
serengil, line 170: instead this, would you please use this

`direction = 1 angle > 0 else -1`

deepface/detectors/DetectorWrapper.py
AndreaLanfranchi: Actually no cause there's no need to rotate when the result of the mod operation is zero.
For example if the user provides an angle of 720 the test `angle==0` would fail but there's no need to rotate in any case as 720 % 360 == 0

deepface/detectors/DetectorWrapper.py
AndreaLanfranchi: Sure thing.

Issue #1056, opened 2024-03-01
Verbosity with error message on verify function 


sanket-tss:
Given an two images are sent to verfiy function in numpy array or base64 format, if in case if the image is not having a face, it does not mentions which image is not having face whether it is img1_path or img2_path
Error message incase base64 is sent: `Face could not be detected in base64 encoded string.`
Error message incase numpy array is sent: `Face could not be detected in numpy array.`

Having detail message whether img1_path or img2_path or both were not having faces will be better.
For now the only way to know which image is not having a face is by doing extract_faces on individual images.

serengil:
Fair enough

serengil:
Closed with PR - https://github.com/serengil/deepface/pull/1059

PR #1057, opened 2024-03-01
returning image name in value error

sanket-tss:
## Tickets

https://github.com/serengil/deepface/issues/1056

### What has been done

With this PR, 
returns argument name img1_path / img2_path, when verify function is called with base64 / numpy array, if face is not found in img1_path or img2_path it return the name _img1_path_ or _img2_path_ with ValueError exception msg.

## How to test

```shell
make lint && make test
```


serengil:
This can be done with simple exception handling in verify instead of changing the interface dramatically

Issue #1052, opened 2024-02-29
Efficiency : wasted computation and instantiation in verify function


AndreaLanfranchi:
Consider [this snippet](https://github.com/serengil/deepface/blob/87f489320578dce4376e1725fa9c53ed32035af7/deepface/modules/verification.py#L112-L134) :

```
    # now we will find the face pair with minimum distance
    for img1_obj in img1_objs:
        img1_content = img1_obj["face"]
        img1_region = img1_obj["facial_area"]
        for img2_obj in img2_objs:
            img2_content = img2_obj["face"]
            img2_region = img2_obj["facial_area"]
            img1_embedding_obj = representation.represent(
                img_path=img1_content,
                model_name=model_name,
                enforce_detection=enforce_detection,
                detector_backend="skip",
                align=align,
                normalization=normalization,
            )

            img2_embedding_obj = representation.represent(
                img_path=img2_content,
                model_name=model_name,
                enforce_detection=enforce_detection,
                detector_backend="skip",
                align=align,
                normalization=normalization,
            )
       [...]

```

`img1_embedding_obj` is recomputed (although with "skip" detector) on every iteration of the inner loop while it should be enough to compute only once per outer loop iteration : in fact is not supposed to change until `img1_obj` steps to the next item in `img1_objs`

serengil:
Good spot!

serengil:
Closed with PR - https://github.com/serengil/deepface/pull/1059

Issue #1051, opened 2024-02-28
Preprocessing.py load_image -> __load_image_from_web


AndreaLanfranchi:
The condition to load an image from web is that `img.startswith("http")`
This however is inaccurate as for example file named `http_image_from_google.png`  (considering the path can be relative) would trigger the load from web (not having any chance to be downloaded) while instead should be loaded from cwd

serengil:
Fair enough

AndreaLanfranchi:
If useful you can extract some hints from my modifications [here](https://github.com/AndreaLanfranchi/deepface/commit/a8b66238b2149006a7ff5fc119db2bee038354a7)

serengil:
Closed with PR - https://github.com/serengil/deepface/pull/1059

Issue #1050, opened 2024-02-28
Preprocessing.py load_image -> load_base64


AndreaLanfranchi:
I believe there _might be_ an inconsistency in the logic of parsing for base64 encoded strings.
The format of input should be something like :

`data:image/jpeg;base64,/9j/4AAQS....`

where the image type is clearly indicated right before the comma.
The parser however does not consider this and, don't know how relevant is, could load any valid image while, for example, in streaming.py only jpg, jpeg, png are considered

serengil:
Right, we should raise an error if type of base64 is not one of these types.

serengil:
Closed with PR - https://github.com/serengil/deepface/pull/1059

Issue #1043, opened 2024-02-26
fastmtcnn sometimes throws TypeError


iAnanich:
Encountered peculiar case when using `fastmtcnn` detector backend - it raises TypeError that is not caught by deepface wrapper.

Happens at `deepface/detectors/FastMtCnn.py:47` :
```python
        if len(detections[0]) > 0:
```

at this moment, some local variables:
```python
detections = (None, [None], None)
```

serengil:
Which version are you using? In the latest version, this is already sorted.

https://github.com/serengil/deepface/blob/master/deepface/detectors/FastMtCnn.py#L30

serengil:
Can you share your input image?

pythonlearner1025:
ran into the same issue. the latest version does not solve it. updating the code to the following does:
  ```if detections[0] is not None and len(detections) > 0:```

serengil:
Closed with PR - https://github.com/serengil/deepface/pull/1059

PR #1059, opened 2024-03-02
Feat task 0203 enhancements

serengil:
## Tickets

- https://github.com/serengil/deepface/issues/1056
- https://github.com/serengil/deepface/issues/1052
- https://github.com/serengil/deepface/issues/1051
- https://github.com/serengil/deepface/issues/1050
- https://github.com/serengil/deepface/issues/1043
- https://github.com/serengil/deepface/issues/1042

### What has been done

With this PR,

1- we were not giving meaningful message if face could not been detected for a given numpy array or base64 encoded image in verify function. this is sorted.
2- in verify function, if one of image pair is having many faces, we were calculating the embedding again and again unncessary.
3- if an image name starts with http, then we were attempting to load it from web
4- in find, we were enforcing images with jpg, jpeg and png but this is not checked for base64 encoded images. to make it compatible, we are now checking type of base64.
5- fastmtcnn detector was throwing error if face could not been detected.
6- stock image of detectors updated.

## How to test

```shell
make lint && make test
```


deepface/modules/preprocessing.py
```diff
@@ -34,7 +34,7 @@ def load_image(img: Union[str, np.ndarray]) -> Tuple[np.ndarray, str]:
         return load_base64(img), "base64 encoded string"
 
     # The image is a url
-    if img.startswith("http"):
+    if img.startswith("http://") or img.startswith("https://"):
```
AndreaLanfranchi, line 37: AFAIK python's string comparison is case sensisitve. Hance this might miss upper case `HTTP://` and all combinations in between. Maybe transform to lower case before ? (as you seem to don't like regexes)

serengil: morning, you are absolutely right. cheers!

creating an issue to follow this.

PR #1063, opened 2024-03-06
Adding a verfication method based on embedding similarity

jake314jake:
## verify_embeddings

```python
def verify_embeddings(
    emb1: np.ndarray,
    emb2: np.ndarray,
    model_name: str = "VGG-Face",
    distance_metric: str = "cosine",
) -> Dict[str, Any]:

```


serengil:
This feature should be embedded into verify function. I do not see any pros of having dedicated function.

PR #1071, opened 2024-03-08
added functionality for dynamic database-picke-file

BrunoMarinhoM:
Now the find function can get also a refresh and updated_files argument so that you can load only once the whole database to a pickle file and, as you are using it, you update the files individualy as needed

### What has been done

With this PR, the find function will be able to work without the need for the whole set of images to be in the database directory/folder which, in cases of a big enough database and/or runing the function in the cloud could lower the prices for persistent memory usage besides making it more secure by not storing the images directly in the computer (but, instead, only its picked representations)

Since no functionality were added to the feature of face-recognition itself, the tests can be made individually by simply adding an image to a data base, turning off the refresh feature and deleting the image and noticing that the function is still working; turning on and adding another file to the path and it to the list of new files and noticing that the only change is that the file will be added to the picke file and the one that was not in the list of updated_images won't be touched.

The documentation was added only within the function file itself, but in case this PR is accepted, I will be also adding it to the readme file as needed. :)

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -235,6 +235,8 @@ def find(
     threshold: Optional[float] = None,
     normalization: str = "base",
     silent: bool = False,
+    refresh_database: bool = True,
+    updated_images: list[str] = []
 ) -> List[pd.DataFrame]:
     """
     Identify individuals in a database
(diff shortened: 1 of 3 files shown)


serengil:
I am currently workin on a branch and it resolved this topic. Will push it in a hour.

BrunoMarinhoM:
Great! Though with the changes I had made the user could select individually which files to modify. Maybe they could compliment each other.

Will look into the new merged version. Thanks for the quick Reply!

serengil:
PR - https://github.com/serengil/deepface/pull/1072 - covers replacements in images already having embeddings.

Issue #1061, opened 2024-03-04
Weak pickled representations consistency in `find`


AndreaLanfranchi:
I believe there might be several improvements in how embeddings for images are stored in pickle file.
Dropping here for reference should you want to address them :

1. [Here a check](https://github.com/serengil/deepface/blob/ad53a9bb2808b895a17b3b76a55b422322b23d03/deepface/modules/recognition.py#L123-L125) is performed to verify the number of columns for the first item matches the number of colums in `df_cols`. Only the number is checked though: not the proper order nor the effective respondence to a data type.
2. The first column on each record of the pickled file is the `identity` which is the combination of the path + file name. This causes any change in the name of the container (e.g. `dataset` -> `haystack`) to completely rebuild the pickle file (as no file is referenced in the pickle file by identity anymore). Probably making the `identity` relative to the position of the pickle file would help. Just a suggestion.
3. Should an image file in dataset be overwritten (keeping the same name) with a totally different content doesn't trigger the reconstruction of the related representation (new and old files are detected only on behalf ot their name)

serengil:
1- We should definetely store column names in pickle

serengil:
2- most probably, we will not do anything for this, because primary key is changed for that case.

serengil:
Closed with PR - https://github.com/serengil/deepface/pull/1072

Issue #1060, opened 2024-03-04
missing uppercase urls


serengil:
@AndreaLanfranchi raised this issue [here](https://github.com/serengil/deepface/pull/1059#discussion_r1510840078)

Currently, we are missing an image path should be downloaded from web if it is uppercase.

Ref: https://github.com/serengil/deepface/blob/master/deepface/modules/preprocessing.py#L37

serengil:
Closed with PR - https://github.com/serengil/deepface/pull/1072

Issue #1049, opened 2024-02-28
[Feature-Request] - Passing embeddings vector in verify function img1_path or img1_path


minhnguyenvan95:
- Call representation function that return embeddings vector
- Verify function should allow passing embedding vector in img1_path or img1_path

serengil:
Fair enough

serengil:
Closed with PR - https://github.com/serengil/deepface/pull/1072

PR #1072, opened 2024-03-08
Feat task 0702 find enhancements

serengil:
## Tickets

- https://github.com/serengil/deepface/issues/1061
- https://github.com/serengil/deepface/issues/1060
- https://github.com/serengil/deepface/issues/1049

### What has been done

With this PR,

1- some improvements done in find function: detect replaced images, store column names in the pickle.
2- do not discard upper case links anymore
3- support pre-calculated embeddings in verify function

## How to test

```shell
make lint && make test
```

### DIFF
--- a/deepface/commons/package_utils.py
+++ b/deepface/commons/package_utils.py
@@ -1,3 +1,6 @@
+# built-in dependencies
+import hashlib
+
 # 3rd party dependencies
 import tensorflow as tf
 
@@ -14,3 +17,16 @@ def get_tf_major_version() -> int:
         major_version (int)
     """
     return int(tf.__version__.split(".", maxsplit=1)[0])
+
+
+def find_hash_of_file(file_path: str) -> str:
+    """
+    Find hash of image file
+    Args:
+        file_path (str): exact image path
+    Returns:
+        hash (str): digest with sha1 algorithm
+    """
+    with open(file_path, "rb") as f:
+        digest = hashlib.sha1(f.read()).hexdigest()
+    return digest
--- a/deepface/modules/recognition.py
+++ b/deepface/modules/recognition.py
@@ -250,21 +250,9 @@ def find(
                     + " after pickle created. Delete the {file_name} and re-run."
                 )
 
-            if distance_metric == "cosine":
-                distance = verification.find_cosine_distance(
-                    source_representation, target_representation
-                )
-            elif distance_metric == "euclidean":
-                distance = verification.find_euclidean_distance(
-                    source_representation, target_representation
-                )
-            elif distance_metric == "euclidean_l2":
-                distance = verification.find_euclidean_distance(
-                    verification.l2_normalize(source_representation),
-                    verification.l2_normalize(target_representation),
-                )
-            else:
-                raise ValueError(f"invalid distance metric passes - {distance_metric}")
+            distance = verification.find_distance(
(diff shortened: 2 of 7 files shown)


deepface/commons/package_utils.py
```diff
@@ -14,3 +17,16 @@ def get_tf_major_version() -> int:
...
+def find_hash_of_file(file_path: str) -> str:
+    """
+    Find hash of image file
+    Args:
+        file_path (str): exact image path
+    Returns:
+        hash (str): digest with sha1 algorithm
+    """
+    with open(file_path, "rb") as f:
+        digest = hashlib.sha1(f.read()).hexdigest()
```
AndreaLanfranchi, line 31: Imo this is quite dangerous.
It attempts to read a whole image file in memory and eventually passes the whole content to the hashing function. In case of (maybe malicious) extremely large image files it could cause OOM (also depending on available memory).
Maybe is better to read and hash the file in chunks.

serengil: Do you recommend to do something like this: https://stackoverflow.com/a/64994148/7846405

deepface/modules/recognition.py
```diff
@@ -120,35 +123,59 @@ def find(
...
+    replaced_images = []
+    for current_representation in representations:
+        identity = current_representation["identity"]
+        if identity in old_images:
+            continue
+        alpha_hash = current_representation["hash"]
+        beta_hash = package_utils.find_hash_of_file(identity)
+        if alpha_hash != beta_hash:
+            logger.debug(f"Even though {identity} represented before, it's replaced later.")
+            replaced_images.append(identity)
```
AndreaLanfranchi, line 159: This is a bit expensive.
You compute the hash of a file here which is eventually discarded.
Then the file with unmatched hash is added to the list of new files which are passed to `_find_bulk_embeddings` where the hash of the file is recomputed once again.

AndreaLanfranchi: I suggest, instead of hashing the whole content of the file, to hash only most immediate (and lighter) properties: name, creation timestamp, last modification timestamp

deepface/commons/package_utils.py
AndreaLanfranchi: Exactly.
As a bonus value for safety I would include a configurable filter in the dataset traverse to exclude in any case files with sizes over a certain threshold: in fact in other parts of the deepface loads in any case the whole file in memory with cv.imread()

deepface/modules/recognition.py
serengil: Liked that approach and convinced. Will do it in another PR soon.

deepface/commons/package_utils.py
serengil: Yeah having file threshold is wise. But will use file properties instead of file itself to hash.

deepface/modules/verification.py
```diff
@@ -175,6 +195,58 @@ def verify(
...
+            img_path=img_path,
+            target_size=target_size,
+            detector_backend=detector_backend,
+            grayscale=False,
+            enforce_detection=enforce_detection,
+            align=align,
+            expand_percentage=expand_percentage,
+        )
+    except ValueError as err:
+        raise ValueError("Exception while processing img1_path") from err
```
sanket-tss, line 230: #1098

PR #1081, opened 2024-03-09
Update requirements.txt

Fadeleke57:
## Tickets

### What has been done

Should be good


serengil:
LGTM

serengil:
Question: should not it be less than tensorflow 2.13?

I mean I suppose this should still work for tf 2.12.1?

serengil:
Realized that dependency is just used for deepface model, so instead of enforcing tf version, I decided to check tf version in that model's initialization in this PR: https://github.com/serengil/deepface/pull/1083

Fadeleke57:
Good solution!

Issue #1120, opened 2024-03-18
Error on verfiy function call: AttributeError: 'tuple' object has no attribute 'shape'


sanket-tss:
PIP versions:

pip install deepface==0.0.83
pip install tensorflow==2.12.0

```python
def convertBase64ToNpArrayBGR(base64_string, img_name):
    decoded_bytes = base64.b64decode(base64_string)
    image_array = np.frombuffer(decoded_bytes, dtype=np.uint8)
    bgr_image = cv2.imdecode(image_array, cv2.IMREAD_COLOR)
    return bgr_image

source_path = "face.png"
target_path = "face.jpg" 

with open(source_path, "rb") as file:
    img1_base64 = str(base64.b64encode(file.read()).decode("utf-8"))

with open(target_path, "rb") as file:
    img2_base64 = str(base64.b64encode(file.read()).decode("utf-8"))


source_np_array = convertBase64ToNpArrayBGR(img1_base64, "src")
target_np_array = convertBase64ToNpArrayBGR(img2_base64, "target")

result = DeepFace.verify(img1_path=source_np_array, img2_path=target_np_array, detector_backend="retinaface", model_name="VGG-Face")
print("match",result["verified"])
print("matchPercentage",round((1-result["distance"]) * 100, 2))
```

Error while calling verify function:
```python

AttributeError                            Traceback (most recent call last)
[<ipython-input-3-e02692f82705>](https://localhost:8080/#) in <cell line: 67>()
     65 target_np_array = convertBase64ToNpArrayBGR(img2_base64, "target")
     66 
---> 67 result = DeepFace.verify(img1_path=source_np_array, img2_path=target_np_array, detector_backend="retinaface", model_name="VGG-Face")
     68 print("match",result["verified"])
     69 print("matchPercentage",round((1-result["distance"]) * 100, 2))

2 frames
[/usr/local/lib/python3.10/dist-packages/deepface/commons/functions.py](https://localhost:8080/#) in extract_faces(img, target_size, detector_backend, grayscale, enforce_detection, align)
    196 
    197     for current_img, current_region, confidence in face_objs:
--> 198         if current_img.shape[0] > 0 and current_img.shape[1] > 0:
    199             if grayscale is True:
    200                 current_img = cv2.cvtColor(current_img, cv2.COLOR_BGR2GRAY)

AttributeError: 'tuple' object has no attribute 'shape'

```

rezakarbasi:
[...]

PR #1124, opened 2024-03-18
add: test script to get response from the api when the image is encoded

rezakarbasi:
- It's an example to address issue #1120

## Tickets

### What has been done

With this PR, ...

## How to test

```shell
make lint && make test
```

### DIFF
--- a/deepface/api/pythontest/apitest.py
+++ b/deepface/api/pythontest/apitest.py
@@ -1,0 +1,55 @@
+# This is a test script to send a POST request to the API endpoint and get the response.
+# Note: API has been brought up using the container of the docker image of the project.
+# Note: this code has been tested on JPG and PNG files only.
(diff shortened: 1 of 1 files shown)


rezakarbasi:
I added a test script to address the issue regarding the sending of an image using base64 encoding

serengil:
Thank you for your contribution, but I think this should not be put into the source.

API has already its unit tests - https://github.com/serengil/deepface/blob/master/tests/test_api.py

rezakarbasi:
yeah. you're right.
But I think you haven't written any tests [here](https://github.com/serengil/deepface/blob/master/tests/test_api.py) to test the encoded images as input. Do you think it'd be suitable to add such a test ?

serengil:
@rezakarbasi yes, please

PR #1127, opened 2024-03-19
feat: an api test added. It evaluates inserting base64 encoded image

rezakarbasi:
…ded input or not

## Tickets

### What has been done

With this PR, ...

## How to test

```shell
make lint && make test
```

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -2,7 +2,6 @@ import unittest
 from deepface.commons.logger import Logger
 from deepface.api.src.app import create_app
 import base64
-import json
 
 
 logger = Logger("tests/test_api.py")
@@ -80,6 +79,7 @@ class TestVerifyEndpoint(unittest.TestCase):
         logger.info("✅ representation api test is done")
 
     def test_represent_encoded(self):
+        image_path = "dataset/img1.jpg"
         with open(image_path, "rb") as image_file:
             encoded_string = "data:image/jpeg;base64," + \
                 base64.b64encode(image_file.read()).decode("utf8")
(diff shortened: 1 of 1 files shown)


tests/test_api.py
```diff
@@ -77,6 +79,33 @@ def test_represent(self):
 
         logger.info("✅ representation api test is done")
 
+    def test_represent_encoded(self):
+        with open(image_path, "rb") as image_file:
```
serengil, line 83: what if image_path?

you may consider to set it to "dataset/img1.jpg"

tests/test_api.py
```diff
@@ -77,6 +79,33 @@ def test_represent(self):
...
+        assert response.status_code == 200
+        result = response.json
+        logger.debug(result)
+        assert result.get("results") is not None
+        assert isinstance(result["results"], list) is True
+        assert len(result["results"]) > 0
+        for i in result["results"]:
+            assert i.get("embedding") is not None
+            assert isinstance(i.get("embedding"), list) is True
+            assert len(i.get("embedding")) == 4096
```
serengil, line 103: Facenet model creates 128 dimensional embeddings, not 4096

tests/test_api.py
```diff
@@ -77,6 +79,33 @@ def test_represent(self):
 
         logger.info("✅ representation api test is done")
 
+    def test_represent_encoded(self):
```
serengil, line 82: you may consider to add similar test case for numpy array input

tests/test_api.py
rezakarbasi: Sure that's the main problem

tests/test_api.py
rezakarbasi: Sure, I'll write it till end of the day

serengil: no problem, you can create a new PR - I am approving this

serengil:
LGTM

PR #1145, opened 2024-03-26
FastMtCnn Use of Gpu for Face Detection

aku-ato:
This pull request introduces an enhancement in the face detection component of our project by improving device compatibility and efficiency. The proposed changes involve dynamically assigning the computational device for the face detection model based on the availability of GPU resources.

**Key Changes:**
- The device selection process is now dynamic, allowing the model to utilize GPU resources when available, falling back to CPU otherwise. This is achieved by checking `torch.cuda.is_available()` and setting the device accordingly.
- This change ensures better performance on systems equipped with GPUs while maintaining compatibility with systems that only have CPUs.

**Modified Code:**
```python
import torch

# Dynamically set the device based on availability
device = torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')
face_detector = fast_mtcnn(device=device)
```

**Benefits:**
- Improved performance on GPU-enabled systems without compromising functionality on CPU-only systems.
- Enhanced efficiency and resource utilization, potentially leading to faster face detection in applications.

I believe this modification will make our face detection more robust and adaptable to various hardware configurations. Looking forward to your feedback and suggestions.

Please review and let me know if any further modifications are needed.

### DIFF
--- a/deepface/detectors/FastMtCnn.py
+++ b/deepface/detectors/FastMtCnn.py
@@ -2,6 +2,7 @@ from typing import Any, Union, List
 import cv2
 import numpy as np
 from deepface.models.Detector import Detector, FacialAreaRegion
+import torch
 
 # Link -> https://github.com/timesler/facenet-pytorch
 # Examples https://www.kaggle.com/timesler/guide-to-mtcnn-in-facenet-pytorch
--- a/deepface/detectors/FastMtCnn.py
+++ b/deepface/detectors/FastMtCnn.py
@@ -2,7 +2,6 @@ from typing import Any, Union, List
 import cv2
 import numpy as np
 from deepface.models.Detector import Detector, FacialAreaRegion
-import torch
 
 # Link -> https://github.com/timesler/facenet-pytorch
 # Examples https://www.kaggle.com/timesler/guide-to-mtcnn-in-facenet-pytorch
@@ -63,6 +62,7 @@ class FastMtCnnClient(Detector):
         # this is not a must dependency. do not import it in the global level.
         try:
             from facenet_pytorch import MTCNN as fast_mtcnn
+            import torch
         except ModuleNotFoundError as e:
             raise ImportError(
                 "FastMtcnn is an optional detector, ensure the library is installed."
--- a/deepface/detectors/FastMtCnn.py
+++ b/deepface/detectors/FastMtCnn.py
@@ -68,7 +68,7 @@ class FastMtCnnClient(Detector):
                 "Please install using 'pip install facenet-pytorch' "
             ) from e
 
-        face_detector = fast_mtcnn(device="cpu")
+        face_detector = fast_mtcnn(device="cuda:0")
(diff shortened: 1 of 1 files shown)


deepface/detectors/FastMtCnn.py
```diff
@@ -2,6 +2,7 @@
 import cv2
 import numpy as np
 from deepface.models.Detector import Detector, FacialAreaRegion
+import torch
```
serengil, line 5: this is not a must dependency in deepface. users will have import error if this import is put in global. right?

deepface/detectors/FastMtCnn.py
```diff
@@ -68,7 +69,9 @@ def build_model(self) -> Any:
                 "Please install using 'pip install facenet-pytorch' "
             ) from e
 
-        face_detector = fast_mtcnn(device="cpu")
+        device = torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')
+        face_detector = fast_mtcnn(device=device)
```
serengil, line 73: to import fast_mtcnn, we [enforce](https://github.com/serengil/deepface/blob/master/deepface/detectors/FastMtCnn.py#L64) users to install facenet_pytorch instead of torch. not sure installing facenet_pytorch will install torch, too.

serengil: okay I can confirm that installing facenet-pytorch will install torchvision, and installing torchvision will install torch

```shell
(base) sefik@L-GCCDXDDT:~$ pip show facenet-pytorch
Name: facenet-pytorch
Version: 2.5.3
Summary: Pretrained Pytorch face detection and recognition models
Home-page: https://github.com/timesler/facenet-pytorch
Author: Tim Esler
Author-email: tim.esler@gmail.com
License:
Location: /home/sefik/miniconda3/lib/python3.9/site-packages
Requires: numpy, pillow, requests, torchvision
Required-by:


(base) sefik@L-GCCDXDDT:~$ pip show torchvision
Name: torchvision
Version: 0.15.2
Summary: image and video datasets and models for torch deep learning
Home-page: https://github.com/pytorch/vision
Author: PyTorch Core Team
Author-email: soumith@pytorch.org
License: BSD
Location: /home/sefik/miniconda3/lib/python3.9/site-packages
Requires: numpy, pillow, requests, torch
Required-by: facenet-pytorch, timm, ultralytics
```

But still putting `import torch` to the line 5 will throw exception for users who do not have facenet-pytorch. Because when i run `pip install deepface`, it will not install torch (it is not a must dependency) and when i import deepface as `from deepface import DeepFace`, then torch will be attempted to be imported but it will throw exception because I do not have that dependency. So, this change breaks deepface code as is.

As you can see facenet_pytorch is imported in line 64 instead of global level: https://github.com/serengil/deepface/blob/master/deepface/detectors/FastMtCnn.py#L64

You may consider to import torch in that try except block.

serengil:
LGTM

PR #1146, opened 2024-03-27
fix a potential exception that resize image to zero width

makui1:
Hello, serengil, thanks for you work.
I find an issue when I using your project--there is a potential exception when using find function. I am new to deep learnning so that my solution may not proper.


serengil:
In line 119, image size is already checked and terminated if one dimension is 0.

Issue #1169, opened 2024-04-06
File image type issue


AndreaLanfranchi:
Actually file images allowed to be loaded are constrained by extensions ["jpg", "jpeg", "png"].
This however does not guarantee file contents are actually images of such type or even images at all (a file extension can be easily changed to whatver user wants).

On my personal version I implemented a check on file content type which is more consistent and as a result, just for your knowledge, I discovered not all files from `tests\dataset` are loaded: in fact "img47.jpg" is a WebP image (not a Jpg).

FYI

serengil:
Closed with PR - https://github.com/serengil/deepface/pull/1171

PR #1171, opened 2024-04-07
checking file's content type instead of extension

serengil:
## Tickets

https://github.com/serengil/deepface/issues/1169

### What has been done

With this PR, we are now checking the file's content type instead of file's extension while we are loading facial database.

## How to test

```shell
make lint && make test
```

### DIFF
--- a/deepface/modules/recognition.py
+++ b/deepface/modules/recognition.py
@@ -3,6 +3,7 @@ import os
 import pickle
 from typing import List, Union, Optional, Dict, Any
 import time
+import imghdr
 
 # 3rd party dependencies
 import numpy as np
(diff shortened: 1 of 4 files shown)


deepface/modules/recognition.py
```diff
@@ -3,6 +3,7 @@
 import pickle
 from typing import List, Union, Optional, Dict, Any
 import time
+import imghdr
```
AndreaLanfranchi, line 6: I discourage the use of `imghdr` as already shows "deprecated" and support will be removed soon. Maybe PIL is better

serengil: Good spot

serengil: Recommended way to determine file type with PIL is

```python
from PIL import Image
img = Image.open(filename)
```

But I do not want to load and open file. If this is the way, then I will prefer to use file extension.

AndreaLanfranchi: You have to read carefully the PIL documentation for the open method on image:

> This is a lazy operation; this function identifies the file, but   the file remains open and the actual image data is not read from the file until you try to process the data (or call the :py:meth:`~PIL.Image.Image.load` method).  

Also img.format reads only as many bytes needed to parse the header (pretty much like imghder.what) hence the operation is very cheap.

serengil: Cheers!

PR #1173, opened 2024-04-07
checking file type with PIL

serengil:
imghdr is deprecated

## Tickets

https://github.com/serengil/deepface/issues/1169
- https://github.com/serengil/deepface/pull/1171#discussion_r1554847133

### What has been done

With this PR, file type check is done with PIL because imghdr is deprecated.

## How to test

```shell
make lint && make test
```

PR #1174, opened 2024-04-07
using with clause to open img with pil to guarantee it is closed

serengil:
## Tickets

- https://github.com/serengil/deepface/issues/1169
- https://github.com/serengil/deepface/pull/1173#pullrequestreview-1984887755

### What has been done

With this PR, we added with close while opening file with PIL to guarantee it is closed after processed.

## How to test

```shell
make lint && make test
```

### DIFF
--- a/deepface/modules/preprocessing.py
+++ b/deepface/modules/preprocessing.py
@@ -88,13 +88,12 @@ def load_base64(uri: str) -> np.ndarray:
     encoded_data = encoded_data_parts[1]
     decoded_bytes = base64.b64decode(encoded_data)
 
-    img = Image.open(io.BytesIO(decoded_bytes))
-    file_type = img.format.lower()
-
     # similar to find functionality, we are just considering these extensions
     # content type is safer option than file extension
-    if file_type not in ["jpeg", "png"]:
-        raise ValueError(f"input image can be jpg or png, but it is {file_type}")
+    with Image.open(io.BytesIO(decoded_bytes)) as img:
+        file_type = img.format.lower()
+        if file_type not in ["jpeg", "png"]:
+            raise ValueError(f"input image can be jpg or png, but it is {file_type}")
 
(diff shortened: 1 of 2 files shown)


deepface/modules/preprocessing.py
```diff
@@ -88,13 +88,12 @@ def load_base64(uri: str) -> np.ndarray:
...
     decoded_bytes = base64.b64decode(encoded_data)
 
-    img = Image.open(io.BytesIO(decoded_bytes))
-    file_type = img.format.lower()
-
     # similar to find functionality, we are just considering these extensions
     # content type is safer option than file extension
-    if file_type not in ["jpeg", "png"]:
-        raise ValueError(f"input image can be jpg or png, but it is {file_type}")
+    with Image.open(io.BytesIO(decoded_bytes)) as img:
```
AndreaLanfranchi, line 93: This is actually not needed as in this case we're not opening a file handle: all data is already in memory (decoded_bytes). Hence the auto closure in a `with` block has no effect.

AndreaLanfranchi: Out of curiosity: why are you proxying `decoded_bytes` through `io.BytesIO`
The signature of Image.open allows to pass directly the `decoded_bytes` bytes object

serengil: You mean `Image.open(decoded_bytes)` performs same? TBH, ChatGPT showed me that usage.

serengil: For with block, agree it has no effect but still i believe it is good practice to confirm opened things closed.

AndreaLanfranchi: > You mean `Image.open(decoded_bytes)` performs same? TBH, ChatGPT showed me that usage.

Never ever trust ChatGPT blindly ... I beg you.
It can give you some hints but **never** let AI write code for you unless you have **rock solid certainty** it is correct, reliable and best practice.
Please.

PR #1227, opened 2024-05-03
feat: add Nix flake

a-h:
This PR adds a Nix flake to the repo.

If you checkout the code and run `nix develop` inside the directory, you'll get a shell containing all of the dependencies required to work on deepface.

Running `nix run` inside the directory results in running the API which can be tested with a `curl` command, e.g.:

```
curl -H "Content-Type: application/json" \
  --data '{ "img_path": "/home/adrian/face-images/couple.jpg" }' \
  http://localhost:5000/represent
```

### DIFF
--- a/deepface/detectors/OpenCv.py
+++ b/deepface/detectors/OpenCv.py
@@ -168,6 +168,9 @@ class OpenCvClient(Detector):
         Returns:
             installation_path (str)
         """
+        if os.getenv("OPENCV_PATH", "") != "":
+            return os.getenv("OPENCV_PATH")
+
         opencv_home = cv2.__file__
         folders = opencv_home.split(os.path.sep)[0:-1]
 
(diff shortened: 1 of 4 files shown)


deepface/detectors/OpenCv.py
```diff
@@ -168,6 +168,9 @@ def __get_opencv_path(self) -> str:
         Returns:
             installation_path (str)
         """
+        if os.getenv("OPENCV_PATH", "") != "":
```
serengil, line 171: Prefer to get opencv path dynamically instead of env vars. Would you please revert this?

mtcnn-0001-merge-pr-129.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/setup.py b/setup.py
```
serengil, line 1: shall we move this to a folder instead of a root? otherwise it will make project messy.

serengil:
closed due to inactivity

PR #1229, opened 2024-05-07
fix docstring max_threshold_to_verify in verify method

GobinathAL:
### What has been done

The docstring of the verify method had a incorrect key `max_threshold_to_verify`(probably due to old code). This PR updates it to `threshold` which is what returned by the verify method.

Note:
The [image in README.md](https://raw.githubusercontent.com/serengil/deepface/master/icon/stock-1.jpg) still has the `max_threshold_to_verify` in it which needs to be updated. Let me know if you want to update it since I don't have the project file.


serengil:

PR #1276, opened 2024-07-13
Fix extract faces grayscale

ProgramadorArtificial:
### What has been done

When activating to return the face in grayscale, an error occurs because it tries to convert from BGR to RGB, but the image is in grayscale.

image

## How to test

Fixed to allow returning the face image in grayscale, to test you can use the following command:
```
    img_path = 'image.jpg',
    grayscale = True
)
```


serengil:
LGTM

PR #1278, opened 2024-07-15
Feature/custom output face from extract faces

ProgramadorArtificial:
### What has been done

Changed extract_faces parameters to be possible to return face image (croped) as rgb, bgr or grayscale and flag for normalized. By default, all parameters will return the same as before (RGB and normalized).

## How to test
Change parameters as desired:
```
    img_path = 'image.jpg',
    normalize_face=True,  # True or False
    color_face='bgr'  # rgb, bgr or gray
)
```
After extracting face can save to see the result:
```
Image.fromarray((faces[0]['face']*255).astype(np.uint8)).save('img.png')  # normalize_face == True
Image.fromarray((faces[0]['face']).astype(np.uint8)).save('img.png')  # normalize_face == False
```

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -503,8 +503,8 @@ def extract_faces(
 
         expand_percentage (int): expand detected facial area with a percentage (default is 0).
 
-        grayscale (boolean): Flag to convert the image to grayscale before
-            processing (default is False).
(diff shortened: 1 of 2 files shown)


serengil:
this design causes troubles to users who are using gray scale argument.

i recommend you to add an additional boolean argument to manage rbg or bgr. this should be run when grayscale is false.

PR #1279, opened 2024-07-15
Feature/custom output extract faces

ProgramadorArtificial:
### What has been done

Changed extract_faces parameters to be possible to return face image (croped) as rgb, bgr or grayscale and flag for normalized. By default, all parameters will return the same as before (RGB and normalized).

## How to test
Change parameters as desired:
```
    img_path = 'image.jpg',
    normalize_face=True,  # True or False
    color_face='bgr'  # rgb, bgr or gray
)
```
After extracting face can save to see the result:
```
Image.fromarray((faces[0]['face']*255).astype(np.uint8)).save('img.png')  # normalize_face == True
Image.fromarray((faces[0]['face']).astype(np.uint8)).save('img.png')  # normalize_face == False
```

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -596,8 +596,6 @@ def detectFace(
         detector_backend=detector_backend,
         enforce_detection=enforce_detection,
         align=align,
-        color_face='rgb',
-        normalize_face=True,
     )
     extracted_face = None
     if len(face_objs) > 0:
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -482,6 +482,7 @@ def extract_faces(
     enforce_detection: bool = True,
     align: bool = True,
     expand_percentage: int = 0,
+    grayscale: bool = False,
     color_face: str = 'rgb',
     normalize_face: bool = True,
     anti_spoofing: bool = False,
@@ -503,6 +504,9 @@ def extract_faces(
         align (bool): Flag to enable face alignment (default is True).
 
         expand_percentage (int): expand detected facial area with a percentage (default is 0).
+
+        grayscale (boolean): (Deprecated) Flag to convert the output face image to grayscale
+            (default is False).
 
         color_face (string): Color to return face image output. Options: 'rgb', 'bgr' or 'gray'
             (default is 'rgb').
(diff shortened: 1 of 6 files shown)


ProgramadorArtificial:
image
I recreated the PR... answering here :)
Ok. Can I add in documentation the "grayscale" parameter is deprecated? (keeping it working)

deepface/modules/verification.py
```diff
@@ -239,7 +239,8 @@ def __extract_faces_and_embeddings(
     img_objs = detection.extract_faces(
         img_path=img_path,
         detector_backend=detector_backend,
-        grayscale=False,
+        color_face='rgb',
+        normalize_face=True,
```
serengil, line 243: default values of color face and normalize face are rgb and True already, IMO we should not set them where it is being called if we are setting default values.

this is available also in representation, recognition,

ProgramadorArtificial: Removed them. I added it because grayscale was there.

deepface/modules/representation.py
```diff
@@ -71,7 +71,6 @@ def represent(
         img_objs = detection.extract_faces(
             img_path=img_path,
             detector_backend=detector_backend,
-            grayscale=False,
```
serengil, line 74: restore this please

deepface/modules/recognition.py
```diff
@@ -364,7 +363,6 @@ def __find_bulk_embeddings(
             img_objs = detection.extract_faces(
                 img_path=employee,
                 detector_backend=detector_backend,
-                grayscale=False,
```
serengil, line 367: restore this please

deepface/modules/recognition.py
```diff
@@ -240,7 +240,6 @@ def find(
     source_objs = detection.extract_faces(
         img_path=img_path,
         detector_backend=detector_backend,
-        grayscale=False,
```
serengil, line 243: restore this please

deepface/DeepFace.py
```diff
@@ -586,7 +596,8 @@ def detectFace(
         detector_backend=detector_backend,
         enforce_detection=enforce_detection,
         align=align,
-        grayscale=False,
```
serengil, line 589: restore this please

deepface/modules/demography.py
```diff
@@ -123,7 +123,6 @@ def analyze(
     img_objs = detection.extract_faces(
         img_path=img_path,
         detector_backend=detector_backend,
-        grayscale=False,
```
serengil, line 126: restore this please

PR #1291, opened 2024-08-02
Update Logger singleton creation, add test and docstring

kremnik:
### What has been done

With this PR, just made a few tweaks to the way we create the singleton instance of the Logger class.

1. The check for the existence of the class instance has been moved from the global dictionary to the `Logger`'s `__new__` method.
2. The function `get_singletonish_logger()` has been replaced with a simple instantiation by calling `Logger()`.
3. A test (in the file `test_singleton.py`) has been added to verify that the created instances have the same memory address.
4. A brief docstring has been added to the Logger class.
5. `.vscode` folder has been added to `.gitignore`.

## How to test

```shell
make lint && make test
```

or if you want to test singleton creation only:

```shell
python -m pytest tests/test_singleton.py -s --disable-warnings
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@ Pipfile
 Pipfile.lock
 .mypy_cache/
 .idea/
-.vscode/
 deepface.egg-info/
 tests/dataset/*.pkl
 tests/*.ipynb
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ Pipfile
 Pipfile.lock
 .mypy_cache/
 .idea/
+.vscode/
 deepface.egg-info/
 tests/dataset/*.pkl
 tests/*.ipynb
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -16,7 +16,7 @@ import tensorflow as tf
 
 # package dependencies
 from deepface.commons import package_utils, folder_utils
-from deepface.commons import logger as log
+from deepface.commons.logger import Logger
 from deepface.modules import (
     modeling,
     representation,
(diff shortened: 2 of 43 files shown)


.gitignore
```diff
@@ -6,6 +6,7 @@ Pipfile
 Pipfile.lock
 .mypy_cache/
 .idea/
+.vscode/
```
serengil, line 9: this should not be added to gitignore because it stores the configuration all developers should follow.

please remove it.

serengil:
In previous design, we can either create a singleton logger and standalone logger. in case of creating standalone one, module is being set and this can be printed in the log messages. With this design, logger must be singleton. How can be highlighted the triggered module?

.gitignore
kremnik: I understand the importance of consistent configurations across different environments.

But I think keeping the `.vscode` folder in the repository can lead to 2 potential issues:

1. The .vscode folder often contains individual developer settings, which might conflict with personal preferences or setups.
2. Sometimes, `.vscode/*` files may include sensitive information specific to a developer's local environment.

To achieve consistency, it is possible to save essential settings in `.vscode/settings.json` while ignoring other files from this folder. 

Let me know your thoughts!

serengil: current .vscode in the source code does not have any sensitive information. still thinking it is important to have it.

kremnik:
I haven't seen the `Logger` being used as a non-singleton in the current design, but it's possible I may have missed it. If having a standalone `Logger` with module-specific messages is needed, I can add argument checking when the Logger is instantiated using `__new__` method. If it includes a "`module`" parameter, it creates a new standalone `Logger` instead of a singleton.
Let me know if this approach works for you!

serengil:
@kremnik never mind the stand alone thing. That feature is available but not being used at all.

It is good to go if you restore vscode folder.

serengil:

PR #1294, opened 2024-08-03
few improvements

light-and-ray:
### What has been done

I use this library and found few things that I need. Maybe you can merge.

1. Disable recursive `db_path` walk in find function
2. Remove `Input image must not have non-english characters` error. In my system all works fine, idk why you've added this. If it's really important, you can make it a warning or optional. Definitely should be disablable 
3. Interupt searching by pressing ctrl+C if `DEEPFACE_KEYBOARD_INTERRUPT` env variable is set. Useful for big searches - just press ctrl+c and thanks to caching, you can get results now, and continue the search later

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -270,6 +270,7 @@ def find(
     silent: bool = False,
     refresh_database: bool = True,
     anti_spoofing: bool = False,
+    recursive: bool = True,
 ) -> List[pd.DataFrame]:
     """
     Identify individuals in a database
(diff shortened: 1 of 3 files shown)


serengil:
1- having 2 branches for recursive or not is not a good design. IMO, going with recursive one is better.
2- Some versions of opencv throws exception if non-english characters available in the file name. You can find this in the closed issues. It was working on my environment, too. However, this doesn't avoid to see issues in repo. So, this behavior mustn't be optional.
3- may be helpful

So, it seems better not to merge this PR. If you create a new PR for the 3rd bullet, then I will be happy to merge.

serengil:
1- You should organize folder for this task. That is not a problem from deepface's perspective.
2- Because if someone is having that issue, then an issue will be raised and I have to respond them all.

PR #1299, opened 2024-08-04
Distances

kremnik:
### What has been done

With this PR, some little optimizations of distance calculations have been done:

1. Using `np.linalg.norm` for norm and Euclidean distance calculations
2. Adding zero denominator checks in the `l2_normalize` function

## How to test

```shell
make lint && make test
```


serengil:
old `find_euclidean_distance` function returns 0.9086272014052611 for dataset/img1.jpg and dataset/img2.jpg whereas new function returns 0.908627201405261

[thresholds](https://github.com/serengil/deepface/blob/master/deepface/modules/verification.py#L371) are tuned for the old function. i am not confident to merge this to cause any trouble.

kremnik:
@serengil The values are extremely close, with the difference appearing in the 15th decimal digit. Since the thresholds precision is set to 2 decimal digits, so small difference should not affect the overall results.

serengil:
yeah you are totally right, it is my bad.

serengil:
TLDR: `np.linalg.norm(a - b)` is equivalent to `np.sqrt(np.sum(np.multiply(a - b, a - b)))`. Both methods compute the Euclidean distance between two vectors a and b.

serengil:
@kremnik LGTM and merged. Thank you for your contribution.

Issue #1301, opened 2024-08-07
[FEATURE]: Refactor Filesystem Operations to Use pathlib Instead of os.path


kremnik:
### Description

I would like to suggest refactoring the codebase to utilize the `pathlib` library for all filesystem operations instead of `os.path`. This change will provide improved readability, code consistency, and platform independence.
If this proposal is approved, I could work on it.

kremnik:
@serengil Yes, of course.

1. In the codebase, both `pathlib` and `os.path` are used. It would be better to use just one library consistently, especially in the files `deepface/commons/folder_utils.py` and `deepface/commons/image_utils.py`;
2. Constructing paths like this `output = home + "/.deepface/weights/" + file_name` (there are a lot of cases of such path constructing style in the codebase) is not safe due to platform-dependent differences in path separators (`/` and `\`);
3. The code in `deepface/commons/constant.py` could be rewritten in a more readable way;
4. There is inconsistency in how existence checks are performed. Sometimes it is written as `if os.path.exists(exact_file_path) is False:` and other times as `if not os.path.exists(weights_path):`. The same inconsistency exists with `os.path.isfile`;
5. There is inconsistent use of `os.path.join` and path concatenation, as mentioned in point 2.

Of course, these points are not crucial, and the code is effective as it stands. However, bringing file operations to a consistent style would improve the code even further.

serengil:
1- we are doing all file related things with `os.path`. Pathlib is just being used to find the home folder, and given image is Path object (this is added with a PR from community, not really required feature). So, **it would be better to give up Pathlib instead of os.path**.
2- constructing paths like `output = home + "/.deepface/weights/" + file_name` makes java programs platform dependent. In python, this is not true. Even in windows, you have to use slash not backslash. If you say this usage is platform dependent, then it wouldn't work in windows but it is working. if you say that using join is more readable, that is fine but i disagree with platform independence thing.
3- don't know how you will do it better
4- I prefer to use 1st usage when I am coding, 2nd usage is most probably coming from a community PR. But that is still totally fine.
5- I am not a fan of join, prefer using slashes when I am coding. These are coming from community PRs.

So, PRs are more than welcome!

kremnik:
1. I agree with you; replacing `Pathlib` with `os.path` will also result in consistent code;
[...]

PR #1305, opened 2024-08-12
Filesystem

kremnik:
## Tickets

https://github.com/serengil/deepface/issues/1301

### What has been done

With this PR, path handling has been updated. In the issue mentioned above, there was a discussion about `pathlib` versus `os.path`. Although `pathlib` offers more advantages, `os.path` is widely used in this project, so the path handling code was updated using `os.path`.

## How to test
Remove the `~/.deepface` folder and run the tests to ensure that all the required models and weights are downloaded and saved correctly.
```shell
make lint && make test
```
Run the tests again to ensure that models and weights are not downloaded again and are loaded correctly.
```shell
make test
```


serengil:

PR #1306, opened 2024-08-13
Update internal db storage

kremnik:
### What has been done

With this PR, the performance of the `in` and `update` operations has been improved by replacing `list` with `set` in the `recognition` module.

## How to test
```shell
make lint && make test
```

### DIFF
--- a/deepface/modules/recognition.py
+++ b/deepface/modules/recognition.py
@@ -1,7 +1,7 @@
 # built-in dependencies
 import os
 import pickle
-from typing import List, Union, Optional, Dict, Any
+from typing import List, Union, Optional, Dict, Any, Set
 import time
 
 # 3rd party dependencies
@@ -141,7 +141,7 @@ def find(
 
     # check each item of representations list has required keys
     for i, current_representation in enumerate(representations):
-        missing_keys = list(set(df_cols) - set(current_representation.keys()))
+        missing_keys = set(df_cols) - set(current_representation.keys())
         if len(missing_keys) > 0:
             raise ValueError(
                 f"{i}-th item does not have some required keys - {missing_keys}."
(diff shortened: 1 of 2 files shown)


deepface/modules/recognition.py
```diff
@@ -194,8 +192,8 @@ def find(
         )
 
     # append replaced images into both old and new images. these will be dropped and re-added.
-    new_images = new_images + replaced_images
-    old_images = old_images + replaced_images
+    new_images.update(replaced_images)
```
serengil, line 195: suppose that `refresh_database  = False`, new_images and old_images are not initialized anymore in line 163 and 164. Line 195 and 196 will throw expception in that case.

Additionally, we should have an unit test for refresh_database = False

serengil:
Secondly, do we have any stats about the improvement? The question is that will it be worth?

kremnik:
@serengil There are three reasons: 
1. Converting `set` to `list` is time consumption, but not neccessary operation;
2. There are a few places when membership checking occurs: [here](https://github.com/serengil/deepface/blob/ff03dc6ece5bc5b38536228766d94b873087d4ed/deepface/modules/recognition.py#L179C1-L182C25) and [here](https://github.com/serengil/deepface/blob/ff03dc6ece5bc5b38536228766d94b873087d4ed/deepface/modules/recognition.py#L200C1-L203C32). If `old_images` includes a huge amount of images (`n`), as well as `representations` includes another huge amount of images (`m`), total time complexity will be `O(nm)` if `old_images` is a `list`. Instead, if `old_images` is a `set`, total time complexity will be `O(m)`;
3. It is guaranteed that there is no re-processing of the same images.

serengil:
LGTM

PR #1309, opened 2024-08-16
Verification

kremnik:
### What has been done

With this PR:
1. The calculation of distances between faces has been updated to optimize memory consumption;
2. Duplicate code for obtaining face embeddings has been moved into a single function.

## How to test
```shell
make lint && make test
```


deepface/modules/verification.py
```diff
@@ -105,79 +105,47 @@ def verify(
...
-                detector_backend=detector_backend,
-                enforce_detection=enforce_detection,
-                align=align,
-                expand_percentage=expand_percentage,
-                normalization=normalization,
-                anti_spoofing=anti_spoofing,
-            )
-        except ValueError as err:
-            raise ValueError("Exception while processing img2_path") from err
+    def extract_faces_from_img(img_path, index=1):
```
serengil, line 108: 1- this function finds embeddings and facial areas. we should give the function a proper name.
2- please add a docstring to this function
3- types of this function are missing. img_path: Union[str, np.ndarray] and index: int. Besides, should we really set a default value to index?
4- return type of this function is missing. it should return Tuple[List[List[float]], List[dict]]

kremnik: I thought that since the function is internal and auxiliary, it wasn't necessary to format it according to the function formatting rules. I'll fix it without any problem.

deepface/modules/verification.py
```diff
@@ -188,21 +156,18 @@ def verify(
...
+            if distance < min_distance:
+                min_distance, min_idx, min_idy = distance, idx, idy
 
     # find the face pair with minimum distance
     threshold = threshold or find_threshold(model_name, distance_metric)
-    min_index = np.argmin(distances)
-    distance = float(distances[min_index])  # best distance
-    facial_areas = facial_areas[min_index]
+    distance = float(min_distance)
+    facial_areas = (no_facial_area, no_facial_area) if None in (min_idx, min_idy) else \
```
serengil, line 169: what if img1_facial_areas[min_idx] is None and img2_facial_areas[min_idy] is not None?

In previous design, we were setting no_facial_area's values to 1st item, and img2_facial_areas[min_idy]'s value to 2nd item.

In your design, no_facial_area's values are not set, instead you are setting None. Am I right?

deepface/modules/verification.py
serengil: Of course, it is not necessary. Still, I strongly recommend to have type hinting.

deepface/modules/verification.py
serengil: I think the following code snippet will resolve that issue

```python
facial_areas = (
	no_facial_area if min_idx is None else img1_facial_areas[min_idx] or no_facial_area,
	no_facial_area if min_idy is None else img2_facial_areas[min_idy] or no_facial_area,
)
```

serengil:
Very good spot btw, I don't like that repeating code and have been planning to fix it. You acted faster than me. Cheers.

serengil:
Linting failed

```
************* Module deepface.modules.verification
deepface/modules/verification.py:114:0: C0301: Line too long (107/100) (line-too-long)
pylint: Command line or configuration file:1: UserWarning: 'Exception' is not a proper value for the 'overgeneral-exceptions' option. Use fully qualified name (maybe 'builtins.Exception' ?) instead. This will cease to be checked at runtime in 3.1.0.
deepface/modules/verification.py:120:0: C0301: Line too long (101/100) (line-too-long)

-----------------------------------
Your code has been rated at 9.99/10
```

deepface/modules/verification.py
kremnik: As I understand it, `img1_facial_areas[min_idx]` and `img2_facial_areas[min_idy]` will never be `None`, because if a face is not found, the area of the entire image is still returned. Therefore, the code from the previous design, `img1_facial_areas[idx] or no_facial_area`, will never return `no_facial_area`.

That's why I thought that if we don't find faces in at least one of the photos, we shouldn't display anything at all.

The code you suggested really solves the problem, thank you!

kremnik: Maybe we don't need `or no_facial_area` in the end of statements? Because `no_facial_area if min_idx is None else img1_facial_areas[min_idx]` will return `Truthy` anyway.
```Python
facial_areas = (
	no_facial_area if min_idx is None else img1_facial_areas[min_idx],
	no_facial_area if min_idy is None else img2_facial_areas[min_idy],
)
```

serengil: Actually, img1_facial_areas[idx] can be None from this line

https://github.com/serengil/deepface/blob/master/deepface/modules/verification.py#L166

serengil: No, the problem occurs when min_idx is 0 but img1_facial_areas[0] is None. We still need to return the dictionary, but the code snippet you recommended returns None.

serengil:
Would you please add an unit test for facial areas checking to https://github.com/serengil/deepface/blob/master/tests/test_verify.py#L123 ?

```python
assert result["facial_areas"]["img1"] is not None
assert result["facial_areas"]["img2"] is not None
```

deepface/modules/verification.py
kremnik: Yes, you are right. I didn't mention, that `img1_facial_areas[idx]` can return `None`. Sorry, my bad.

What do you think, maybe, it is better to do something like this?
```python
img_facial_areas = [no_facial_area]
```
instead of 
```python
img_facial_areas = [None]
```

serengil: yeah, this will also work

i hope this is the only point facial area can be none - https://github.com/serengil/deepface/blob/master/deepface/modules/verification.py#L166

deepface/modules/verification.py
```diff
@@ -188,21 +114,88 @@ def verify(
...
+            # given image is already pre-calculated embedding
+            if not all(isinstance(dim, float) for dim in img_path):
+                raise ValueError(
+                    f"When passing img{index}_path as a list,"
+                    " ensure that all its items are of type float."
+                )
+
+            if silent is False:
+                logger.warn(
+                    "You passed 1st image as pre-calculated embeddings."
```
serengil, line 153: 1st image term should not be static. Would you please get it from index arg?

deepface/modules/verification.py
```diff
@@ -188,21 +114,88 @@ def verify(
...
+                    "Please ensure that embeddings have been calculated"
+                    f" for the {model_name} model."
+                )
+
+            if len(img_path) != dims:
+                raise ValueError(
+                    f"embeddings of {model_name} should have {dims} dimensions,"
+                    f" but it has {len(img_path)} dimensions input"
```
serengil, line 161: shall we mention the index here as well?

serengil:
LGTM, I added a couple of minor comments but I can add them later. Cheers!

Issue #1318, opened 2024-08-23
[FEATURE]: Improve represent performance


Circuit8:
### Description

Performance is very slow when aligning faces when using the `represent` API. Please see my PR here: https://github.com/serengil/deepface/pull/1319

serengil:
What do you recommend then?

Circuit8:
Please see my PR: https://github.com/serengil/deepface/pull/1319

PR #1319, opened 2024-08-23
Represent optimizations

Circuit8:
https://github.com/serengil/deepface/issues/1318

I've discovered that when using the `represent` api, the align phase is the slowest part, and was causing a bottleneck for my usecase. 

The slow part was actually in aligning all the faces in the image. I have created a PR that speeds this up by using cv2 to rotate the face image slice rather than converting it into a PIL image. This speeds up the process dramatically.

I have also ensured that the `max_faces` parameter is passed down into the `detect_faces` function so that it only aligns the number of faces it needs to, rather than aligning them all and discarding the ones we dont need.

Together these two changes provide roughly an order of magnitude improvement in terms of speed when calling `represent`.

I just have one question though, is the way I'm aligning the faces with cv2 optimal / correct, or is there something different with the implementation compared to PIL? I don't have much experience with python image manipulation, but it seems to work well for my use case.

### DIFF
--- a/deepface/modules/detection.py
+++ b/deepface/modules/detection.py
@@ -2,9 +2,9 @@
 from typing import Any, Dict, List, Tuple, Union, Optional
 
 # 3rd part dependencies
+from heapq import nlargest
 import numpy as np
 import cv2
-from PIL import Image
 
 # project dependencies
 from deepface.modules import modeling
@@ -179,7 +179,9 @@ def extract_faces(
 
 
 def detect_faces(
-    detector_backend: str, img: np.ndarray, align: bool = True, expand_percentage: int = 0, max_faces: Optional[int] = None
+    detector_backend: str, img: np.ndarray,
+    align: bool = True, expand_percentage: int = 0,
+    max_faces: Optional[int] = None
 ) -> List[DetectedFace]:
     """
     Detect face(s) from a given image
@@ -205,7 +207,6 @@ def detect_faces(
         - confidence (float): The confidence score associated with the detected face.
     """
     height, width, _ = img.shape
-    
     face_detector: Detector = modeling.build_model(
         task="face_detector", model_name=detector_backend
     )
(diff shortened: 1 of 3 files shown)


serengil:
Seems this change broken the code

https://github.com/serengil/deepface/blob/master/tests/test_verify.py#L69

This unit test got 85.33 accuracy score in the pipeline - https://github.com/serengil/deepface/actions/runs/10528897772/job/29175534942?pr=1319#step:5:1264

On the other hand, before this change it was getting 99.33% score - https://github.com/serengil/deepface/actions/runs/10522084149/job/29154000820#step:5:968

Circuit8:
Ah yes I see, looks like its only working correctly with Facenet512 which is what I'm using. I'll have a look to see if I can make it work for the others aswell.

Circuit8:
@serengil my latest change should have the tests passing.

serengil:
Passed unit tests but failed at linting

deepface/modules/detection.py
```diff
@@ -233,6 +237,17 @@ def detect_faces(
...
 
+    if max_faces is not None and max_faces < len(facial_areas):
+        # sort as largest facial areas come first
+        facial_areas = sorted(
+            facial_areas,
+            key=lambda facial_area: facial_area.w * facial_area.h,
+            reverse=True,
+        )
+        # discard rest of the items
+        facial_areas = facial_areas[0:max_faces]
```
kremnik, line 247: I think this code could be more memory and computation efficient by using the heap structure:
```python
from heapq import nlargest

if max_faces is not None and max_faces < len(facial_areas):
    facial_areas = nlargest(
        max_faces,
        facial_areas,
        key=lambda facial_area: facial_area.w * facial_area.h
    )
```
The time complexity of this approach is `O(nlogk)` (where `n` is `len(facial_areas)` and `k` is `max_faces`), compared to sorting, which is `O(nlogn)`. Additionally, the memory consumption here is `O(k)` instead of `O(n)`. 
This optimization could be important when the number of faces is much greater than `max_faces`.

deepface/modules/detection.py
```diff
@@ -247,62 +246,62 @@ def detect_faces(
...
-
-        result = DetectedFace(
-            img=detected_face,
-            facial_area=FacialAreaRegion(
-                x=x, y=y, h=h, w=w, confidence=confidence, left_eye=left_eye, right_eye=right_eye
-            ),
-            confidence=confidence,
-        )
-        results.append(result)
+        results.append(expand_and_align_face(facial_area=facial_area, img=img, align=align, expand_percentage=expand_percentage, width_border=width_border, height_border=height_border))
```
kremnik, line 252: I also think this code could be a bit faster using list comprehension:
```python
results = [
    expand_and_align_face(
        facial_area=facial_area,
        img=img,
        align=align,
        expand_percentage=expand_percentage,
        width_border=width_border,
        height_border=height_border
    )
    for facial_area in facial_areas
]
```
List comprehensions are faster than a loop with `append` because the list is allocated only once, rather than being dynamically resized.

Circuit8:
> Besides, would you please share the processing time for before and after?
> 
> Just wondered.

The original PIL implementation was taking around 0.05s per face, the new cv2 version takes between 0.7-1.0s per face.

deepface/modules/detection.py
Circuit8: OK good idea, changed.

Circuit8:
@serengil made a new commit including your suggestions and with the linter passing.

serengil:
LGTM

PR #1321, opened 2024-08-27
Optimize ssd detector

kremnik:
## What has been done

With this PR, `pandas` has been removed from the SSD detector script. All computations are now performed in matrix form using `numpy`, which speeds up calculations and reduces memory consumption.
Additionally, a test for SSD was added to `test_extract_faces.py`, and a bug related to checking for a black image was fixed: `cv2.dnn.blobFromImage` expects `np.uint8` as input, while `np.zeros` by default generates `np.float64`.

## How to test
```shell
make lint && make test
```

### DIFF
--- a/deepface/models/face_detection/Ssd.py
+++ b/deepface/models/face_detection/Ssd.py
@@ -2,7 +2,6 @@ from typing import List
 import os
 import gdown
 import cv2
-import pandas as pd
 import numpy as np
 from deepface.models.face_detection import OpenCv
 from deepface.commons import folder_utils
@@ -50,11 +49,10 @@ class SsdClient(Detector):
                 + "You can install it as pip install opencv-contrib-python."
             ) from err
 
-        detector = {}
-        detector["face_detector"] = face_detector
-        detector["opencv_module"] = OpenCv.OpenCvClient()
-
-        return detector
+        return {
+            "face_detector": face_detector,
+            "opencv_module": OpenCv.OpenCvClient()
+        }
 
     def detect_faces(self, img: np.ndarray) -> List[FacialAreaRegion]:
         """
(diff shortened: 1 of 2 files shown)


serengil:
IMO, implementation with pandas is much cleaner and readable.

no idea what is going on these lines

```python
        faces = detections[0][0]
        faces = faces[(faces[:, 1] == 1) & (faces[:, 2] >= 0.90)]
        faces[:, 3:7] = np.int32(faces[:, 3:7] * 300)
        faces[:, 3:7] = np.int32(faces[:, 3:7] * [aspect_ratio_x, aspect_ratio_y, aspect_ratio_x, aspect_ratio_y])
        faces[:, 5:7] -= faces[:, 3:5]
```

PS: would you please raise a ticket before creating a PR to discuss first, otherwise a PR may not be merged, and your effort goes wasted.

kremnik:
@serengil Ok, I will try to make it more readable.
The purpose of this PR is to avoid using slow `pandas` when possible. Also, this is not a piece of code that users actively edit.

serengil:
TBH, I am satisfied with pandas because it comes with readability. No matter users aren't touching this.

kremnik:
@serengil `pandas` is great for data analysis and some data pre-processing, but it can be quite slow for production purposes, especially when filtering.
For example, the following code:
```python
detections_df = detections_df[detections_df["is_face"] == 1]  # 0: background, 1: face
detections_df = detections_df[detections_df["confidence"] >= 0.90]
```
takes around 26ms on `selfie-many-people.jpg` on my cpu.

However, this code: 
```python
faces = faces[(faces[:, 1] == 1) & (faces[:, 2] >= 0.90)]
```
takes around 16μs on the same photo and the same cpu.

serengil:
Then, you convinced me to switch :) 

But please make it readable with comments at least. Because I have to fight with issues and the understanding the code not written by myself is hard.

Storing index keys to variables would be good.

e.g. idx_face = 1 and faces[:, idx_face] == 1

serengil:
LGTM

PR #1334, opened 2024-09-02
Improvement: check image exists first

kremnik:
## What has been done

With this PR, image existence checking has been added in `recognition.py`. The reason for this change is to solve the situtation that occurs when attempting to find similar faces using the `find` function. Previously, the function first collected embeddings from the database before searching for faces in the target image. If the path to the target image was incorrect or the image did not exist, the exception would only be raised **after** the database scan, which could be time-consuming.

## How to test
```shell
make lint && make test
```


serengil:
LGTM

Issue #1332, opened 2024-09-01
[FEATURE]: Add an optional flag to the "find" function in recognition.py to return an ndarray.


kremnik:
### Description

Currently, the `find` function in `recognition.py` file returns the information about faces in a Pandas `dataframe`, which is actually very convenient. However, when using this function on a large photo database, the `dataframe` structure can significantly slow down the finding process, particularly for operations like filtering and sorting.
I suggest adding an optioanl flag to the function that allows it to process and return raw `ndarray` instead of creating a `dataframe`.

serengil:
it would be a good feature ONLY IF its default behaviour isn't changed. We don't want to change the default interface of find function.

kremnik:
I agree with you, the default behavior should not be changed.

serengil:
appreciate @kremnik as always

PR #1351, opened 2024-09-30
Recognition find batched

kremnik:
## Tickets

https://github.com/serengil/deepface/issues/1332

### What has been done

With this PR, the `find_batched` method was developed as a batched version of the `find` method. The main difference of that method is using numpy arrays instead of `pd.DataFrame` and using batched distance calculation instead of element-wise distance calculation. Also, the function returns array of array of dicts (where `array[:]` is source faces and `array[:][:]` is founded faces corresponding to each source face) instead of array of DataFrames.

Also, the tests for the `find_batched` method was developed in the `tests/test_find_batched.py`.

Based on the experiments conducted on large face databases, which include photos with multiple faces, the `find_batched` method achieves a speedup of more than 50% compared to the `find` method.

## How to test
```shell
make lint && make test
```

### DIFF
--- a/deepface/modules/recognition.py
+++ b/deepface/modules/recognition.py
@@ -505,7 +505,7 @@ def find_batched(
     """
     embeddings_list = []
     valid_mask = []
-    other_keys = set()
+    metadata = set()
 
     for item in representations:
         emb = item.get('embedding')
@@ -516,18 +516,18 @@ def find_batched(
             embeddings_list.append(np.zeros_like(representations[0]['embedding']))
             valid_mask.append(False)
 
-        other_keys.update(item.keys())
+        metadata.update(item.keys())
 
     # remove embedding key from other keys
-    other_keys.discard('embedding')
-    other_keys = list(other_keys)
+    metadata.discard('embedding')
+    metadata = list(metadata)
 
     embeddings = np.array(embeddings_list) # (N, D)
     valid_mask = np.array(valid_mask) # (N,)
 
     data = {
         key: np.array([item.get(key, None) for item in representations])
-        for key in other_keys
+        for key in metadata
     }
 
     target_embeddings = []
--- a/deepface/modules/recognition.py
+++ b/deepface/modules/recognition.py
@@ -453,7 +453,7 @@ def find_batched(
     anti_spoofing: bool = False,
 ) -> List[List[Dict[str, Any]]]:
     """
-    Perform batched face recognition by comparing source face embeddingswith a set of
+    Perform batched face recognition by comparing source face embeddings with a set of
     target embeddings. It calculates pairwise distances between the source and target
     embeddings using the specified distance metric.
     The function uses batch processing for efficient computation of distances.
@@ -549,6 +549,7 @@ def find_batched(
             align=align,
             normalization=normalization,
         )
+        # it is safe to access 0 index because we already fed detected face to represent function
         target_representation = target_embedding_obj[0]["embedding"]
 
         target_embeddings.append(target_representation)
@@ -566,21 +567,6 @@ def find_batched(
         'source_h': np.array([region['h'] for region in source_regions]),
     }
 
-    def l2_normalize(
-        x: np.ndarray, axis: int = 1, epsilon: float = 1e-10
-    ) -> np.ndarray:
-        """
-        Normalize input vectors along a specified axis using L2 normalization
-        Args:
-            x (np.ndarray): input array
-            axis (int): axis along which to normalize
-            epsilon (float): small value to avoid division by zero
-        Returns:
-            np.ndarray: L2-normalized array of the same shape as input
-        """
-        norm = np.linalg.norm(x, axis=axis, keepdims=True)
-        return x / (norm + epsilon)
-
     def find_cosine_distance_batch(
         embeddings: np.ndarray, target_embeddings: np.ndarray
     ) -> np.ndarray:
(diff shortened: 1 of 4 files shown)


deepface/modules/recognition.py
```diff
@@ -415,3 +440,233 @@ def __find_bulk_embeddings(
...
+    target_embeddings = np.array(target_embeddings) # (M, D)
+    target_thresholds = np.array(target_thresholds) # (M,)
+    source_regions_arr = {
+        'source_x': np.array([region['x'] for region in source_regions]),
+        'source_y': np.array([region['y'] for region in source_regions]),
+        'source_w': np.array([region['w'] for region in source_regions]),
+        'source_h': np.array([region['h'] for region in source_regions]),
+    }
+
+    def l2_normalize(
```
serengil, line 569: we have same function under verification module - https://github.com/serengil/deepface/blob/master/deepface/modules/verification.py

you may consider not to duplicate same function with using it.

tests/test_find_batched.py
```diff
@@ -0,0 +1,99 @@
...
+    assert len(results) > 0
+    for result in results:
+        assert isinstance(result, list)
+
+        found_image_itself = False
+        for face in result:
+            assert isinstance(face, dict)
+            if face["identity"] == img_path:
+                # validate reproducability
+                assert face["distance"] < threshold
```
serengil, line 32: this condition must be satisfied not only for face["identity"] == img_path

you may consider to move this above if condition

tests/test_find_batched.py
```diff
@@ -0,0 +1,99 @@
...
+def test_find_with_exact_path():
+    img_path = os.path.join("dataset", "img1.jpg")
+    results = DeepFace.find(img_path=img_path, db_path="dataset", silent=True, batched=True)
+    assert len(results) > 0
+    for result in results:
+        assert isinstance(result, list)
+
+        found_image_itself = False
+        for face in result:
+            assert isinstance(face, dict)
```
serengil, line 29: you may consider to check required keys in face

deepface/modules/recognition.py
kremnik: The original version doesn't support `axis` and `keepdims` parameters, which is important for batched prosessing.

serengil: won't adding them to the original one be better?

tests/test_find_batched.py
kremnik: As I got, this condition block tests exactly comparing the image with itself. You do the same in `test_find.py`:
```python
# one is img1.jpg itself
identity_df = df[df["identity"] == img_path]
assert identity_df.shape[0] > 0

# validate reproducability
assert identity_df["distance"].values[0] < threshold
```

serengil:
this is a comprehensive PR, need to spend some time on it

deepface/modules/recognition.py
kremnik: No problem, I just thought that batched functions incapsulation is more clean.

deepface/modules/recognition.py
```diff
@@ -415,3 +440,233 @@ def __find_bulk_embeddings(
...
+
+        target_embedding_obj = representation.represent(
+            img_path=source_img,
+            model_name=model_name,
+            enforce_detection=enforce_detection,
+            detector_backend="skip",
+            align=align,
+            normalization=normalization,
+        )
+        target_representation = target_embedding_obj[0]["embedding"]
```
serengil, line 552: would you please add a comment here like 

it is safe to access 0 index because we already fed detected face to represent function

deepface/modules/recognition.py
```diff
@@ -415,3 +440,219 @@ def __find_bulk_embeddings(
...
+    ) -> np.ndarray:
+        """
+        Find the cosine distances between batches of embeddings
+        Args:
+            embeddings (np.ndarray): array of shape (N, D)
+            target_embeddings (np.ndarray): array of shape (M, D)
+        Returns:
+            np.ndarray: distance matrix of shape (M, N)
+        """
+        embeddings_norm = verification.l2_normalize(embeddings, axis=1)
```
serengil, line 581: you performed cosine l2 here but we mostly don't apply l2 to cosine

kremnik: L2 normalization is exactly what `np.linalg.norm` do. The `l2_normalize` method is just a wrap around `np.linalg.norm`. Therefore applying `verification.l2_normalize` in the `find_cosine_distance_batch` is the same as applying `np.linalg.norm` in the `find_cosine_distance`, but taking into account that we do it batch-wise:

```python
def find_cosine_distance(
    source_representation: Union[np.ndarray, list], test_representation: Union[np.ndarray, list]
) -> np.float64:
    ...
    a = np.dot(source_representation, test_representation)
    b = np.linalg.norm(source_representation)
    c = np.linalg.norm(test_representation)
    return 1 - a / (b * c)
```

```python
def find_cosine_distance_batch(
    embeddings: np.ndarray, target_embeddings: np.ndarray
) -> np.ndarray:
    ...
    embeddings_norm = verification.l2_normalize(embeddings, axis=1)
    target_embeddings_norm = verification.l2_normalize(target_embeddings, axis=1)
    cosine_similarities = np.dot(target_embeddings_norm, embeddings_norm.T)
    cosine_distances = 1 - cosine_similarities
    return cosine_distances
```

Example to test:
```python
x = np.random.randn(100)
y = np.random.randn(100)
ans1 = find_cosine_distance(x, y)
ans2 = find_cosine_distance_batch([x], [y])
print(np.round(ans1, 10) == np.round(ans2, 10)) # True
```

tests/test_find_batched.py
kremnik: https://github.com/serengil/deepface/pull/1351/commits/3bb317ac63f87624bd8b77d76e919d42edff13c5#diff-c9703fe704ffb9ab295563a197a264d2d938a16b4dd67bacd9cdb5ee26a85b8dR22-R26

Issue #1368, opened 2024-10-16
Get log for any request when running in docker


seavphing:
### Description

When I use deepface, I see that for any incoming request, there is no log to monitor the request, so in the Dockerfile I modified it to use entrypoint and give gunicorn a better log for any request.

PR #1369, opened 2024-10-16
improve Dockerfile to use entrypoint in order to get better log when …

seavphing:
…receive any request

## Tickets

https://github.com/serengil/deepface/issues/1368

### What has been done

With this PR, I have update the Dockerfile to use entrypoint.sh for a better log in order to get easy monitor when the program is running.

## How to test

```shell
Just run the dockerfile and see if the log is display or not
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,4 +55,5 @@ ENV PYTHONUNBUFFERED=1
 # run the app (re-configure port if necessary)
 WORKDIR /app/deepface/api/src
 EXPOSE 5000
-CMD ["gunicorn", "--workers=1", "--timeout=3600", "--bind=0.0.0.0:5000", "app:create_app()"]
+# CMD ["gunicorn", "--workers=1", "--timeout=3600", "--bind=0.0.0.0:5000", "app:create_app()"]
+ENTRYPOINT [ "sh", "entrypoint.sh" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,14 +7,21 @@ LABEL org.opencontainers.image.source https://github.com/serengil/deepface
 RUN mkdir /app
 RUN mkdir /app/deepface
 
+
+
 # -----------------------------------
 # switch to application directory
 WORKDIR /app
 
 # -----------------------------------
 # update image os
-RUN apt-get update
-RUN apt-get install ffmpeg libsm6 libxext6 -y
+# Install system dependencies
+RUN apt-get update && apt-get install -y \
+    ffmpeg \
+    libsm6 \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -55,4 +55,5 @@ ENV PYTHONUNBUFFERED=1
 # run the app (re-configure port if necessary)
 WORKDIR /app/deepface/api/src
 EXPOSE 5000
-CMD ["gunicorn", "--workers=1", "--timeout=3600", "--bind=0.0.0.0:5000", "app:create_app()"]
+# CMD ["gunicorn", "--workers=1", "--timeout=3600", "--bind=0.0.0.0:5000", "app:create_app()"]
+ENTRYPOINT [ "sh", "entrypoint.sh" ]
```
serengil, line 59: shouldn't we copy the entypoint shell scrip into the image?

seavphing: My bad sir, I forgot to do double check and commit the wrong branch, I'll correct it immediately

seavphing: @serengil Hello sir, I already fix and test everything sir.

requirements_local
```diff
@@ -2,5 +2,5 @@ numpy==1.22.3
 pandas==2.0.3
 Pillow==9.0.0
 opencv-python==4.9.0.80
-tensorflow==2.9.0
-keras==2.9.0
+tensorflow==2.13.1
```
serengil, line 5: what is the purpose of these version changes? if it is not a must, i will ask you not to change requirements.

Dockerfile
```diff
@@ -7,14 +7,21 @@ LABEL org.opencontainers.image.source https://github.com/serengil/deepface
...
 # -----------------------------------
 # update image os
-RUN apt-get update
-RUN apt-get install ffmpeg libsm6 libxext6 -y
+# Install system dependencies
+RUN apt-get update && apt-get install -y \
+    ffmpeg \
+    libsm6 \
+    libxext6 \
+    libhdf5-dev \
```
serengil, line 23: what is the purpose of installation of libhdf5-dev?

requirements_local
seavphing: @serengil As I test it sir, when building the image, at pip install for tensorflow version 2.9.0, it can't find that version and it suggest for tensorflow version 2.13.1 sir.

Dockerfile
seavphing: @serengil I install libhdf5-dev to prevent h5py package failed to build sir.

serengil:
LGTM

PR #1370, opened 2024-10-18
Update requirements.txt

cobycloud:
## Tickets

deepface requires unnecessary depth resolution of retina_face upon installation of deepface dependencies

image

### What has been done

tightened the versioning of retina_face so that resolvers only need to go 3 deep

## How to test

Build dev release, test, publish as dev release, encourage small group install locally for feedback before publishing as stable


serengil:
The latest version of retina-face is 0.0.17, I think we should add this into the requirements. Do you mind to do that?

cobycloud:
I concur! 🙏

PR #1372, opened 2024-10-20
Add find_closest_embedding function in verification.py to find the index of the closest embedding to the current embedding from a list of embeddings

YoussefAboelwafa:
Add **find_closest_embedding** function in **verification.py** to find the index of the closest embedding to the current embedding from a list of embeddings if less than a certain threshold.

```python
def find_closest_embedding(
    current_embedding: List[float],
    embeddings_list: List[List[float]],
    distance_metric: str = "cosine",
    threshold: Optional[float] = 0.5,
) -> Optional[int]
```

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -162,6 +162,19 @@ def verify(
         anti_spoofing=anti_spoofing,
     )
 
+def find_closest_embedding(
+    current_embedding: List[float],
+    embeddings_list: List[List[float]],
+    distance_metric: str = "cosine",
+    threshold: Optional[float] = 0.5,
+) -> Optional[int]:
+    
+    return verification.find_closest_embedding(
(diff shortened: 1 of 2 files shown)


LD5563 (approved):

LD5563:
BOM VC JOGO BEM

serengil:
Thank you for your contribution! While the find_closest_embedding function might be useful in some cases, I don’t see a reason to merge it at this time. The function is not being used within the codebase, and users already have access to similar functionality through the verify or find methods of DeepFace, which are more comprehensive and widely adopted. Introducing another function might lead to redundancy without adding significant value.

Please feel free to discuss further if you see a specific use case where this function would be more beneficial than the existing ones.

serengil:
of course you can adopt this but this will not bring any value to most of users

Issue #1101, opened 2024-03-11
Representing a batch of multiple images


kerighan:
Thanks for your helpful library.
I wonder how it is possible to infer the embedding of multiple pictures at once, so as to maximally leverage model parallelism and improve inference speed. From the code, it does not seem obvious

serengil:
You should call represent in a for loop.

No plan to support batches in the short term.

Issue #1069, opened 2024-03-06
Analyze on a batch of images


tomer-erez:
Hey guys and girls. Is it possible to call deepface analyze over a batch of images? A whole folder with multiple images? I want to use it to predict ages of an entire folder without for or while loops.

tomer-erez:
Also if there are tips to speed up inference time i would love to hear it

pyshen-watson:
same question

serengil:
No, you cannot use barches.

Issue #678, opened 2023-02-16
Batch Predictions


swapnika92:
How can we get facial attributes (gender, age, etc.) for multiple images (BATCH PREDICTIONs)?
Currently, "analyze" function builds the model when run for each image individually - leading to longer processing time.

serengil:
you should call the function in a for loop. ml models built with singleton pattern. so, you do not spend time to re-built model in that loop.

milesmcc:
@serengil would it be possible to extend the library to support batch inference when running on GPUs? While the ML model has a singleton pattern and so it isn't re-built each call, it would be helpful to have a method to truly parallelize inference.

If you think this is possible, I'm happy to try and submit a PR myself or pay a bounty to the person who does (though it'd need to be small because I don't have much money haha).

Issue #441, opened 2022-03-16
Question about multiple image analysis


ligujie123:
Hi Sefik! Thanks for sharing the library. I use your library for race detection for about 1000 images. As you suggested in some comments below the YouTube video, I use list to store image names, and in this way, it will not kill much memory. 

My python code is like this:

from deepface import DeepFace
backends = ['opencv', 'ssd', 'dlib', 'mtcnn', 'retinaface', 'mediapipe']
obj = DeepFace.analyze(["testface1.jpg", "testface2.jpg", "testface3.png"], actions = ['race'], enforce_detection = True, detector_backend = backends[3])

My goal is to obtain the predicted race information in obj, which is a dictionary produced by the DeepFace.analyze function. However, for some images (for example, testface2 and testface3), there is no face detected, so there shows error message "Face could not be detected. Please confirm that the picture is a face photo or consider to set enforce_detection param to False." The obj function cannot be obtained. I tried to turn enforce_detection to False but to ensure my result is accurate enough, I hope to turn it to be True. I am wondering if there is a way that I still can get obj, but for the images for which face not detected, it just produce a value like "not detected" instead of terminating the function without any obj produced.

Hope my question is explained clearly.

Thanks!

serengil:
facial_imgs = ["testface1.jpg", "testface2.jpg", "testface3.png"]

Method 1: Set enforce detection the False. In this way, you will not have any trouble. However, accuracy will be down.

`obj = DeepFace.analyze(facial_imgs, actions = ['race'], enforce_detection = False, detector_backend = backends[3])
`

Method 2: Build a for loop. Previous versions of deepface had a memory error in this approach but not you can call it in a for loop. It is totally fine.

```
facial_imgs = ["testface1.jpg", "testface2.jpg", "testface3.png"]

for facial_img in facial_imgs:
   try:
      obj = DeepFace.analyze(facial_img, actions = ['race'], enforce_detection = True, detector_backend = backends[3])
   except:
      print("not detected - " + facial_img)
```

ligujie123:
Thanks a lot! Really helpful suggestions!

CodeOfDuty44:
Hello, thanks for your work. One question: Does using for loop for predicting attributes for batch of images slow down the overall computation, since the model will be loaded and unloaded for every function call?

serengil:
@CodeOfDuty44 no it is singletonish, it is loaded once and be reused from the 2nd iteration

CodeOfDuty44:
[...]

PR #1396, opened 2024-12-06
Feat/batch predict age and gender

NatLee:
## Tickets

https://github.com/serengil/deepface/issues/441
https://github.com/serengil/deepface/issues/678
https://github.com/serengil/deepface/issues/1069

### What has been done

With this PR, new predicts function to support batch predictions.

## How to test

Use this class, user can load model and predict in batch.

```python
class AgeGenderModel():
    """
    Age and gender model
    """
    def __init__(self):
        self.target_size = (224, 224)
        self.age_model, self.gender_model = self.load()

    def load(self) -> Tuple:
        age_model = modeling.build_model(task="facial_attribute", model_name="Age")
        gender_model = modeling.build_model(task="facial_attribute", model_name="Gender")
        return age_model, gender_model

    def process_data(self, data: np.ndarray) -> np.ndarray:
        """
        Process input image data
        """
        img_content = data[:, :, ::-1] # rgb to bgr
        img_content = preprocessing.resize_image(
            img=img_content,
            target_size=self.target_size
        )
        return img_content

    def predict(self, data: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
        """
        Predict age and gender for one input image
        """
        img_content = self.process_data(data)
        age = self.age_model.predict(img_content)
        gender = self.gender_model.predict(img_content)
        return gender, age

    def predicts(self, data: List[np.ndarray]) -> Tuple[List, List]:
        """
        Predict age and gender for batch input images
        """
        img_contents = [self.process_data(img_content) for img_content in data]
        ages = self.age_model.predicts(img_contents)
        genders = self.gender_model.predicts(img_contents)
        return genders, ages 
```

Time Costs for 10 Images:
```
Original for-loop prediction: 1.91108 seconds

Loading split Age and Gender models: 0.00002 seconds
Batch prediction: 0.58243 seconds
Single prediction: 0.03749 seconds
```

Here's my test script:

```python
from deepface import DeepFace
import cv2
import time
# Load one face
img = cv2.imread("./test.png")

# Make it as a list
imgs = [img] * 10

# For-loop predict
start = time.time()
print("=====For-loop Predict=====")
for img in imgs:
    objs = DeepFace.analyze(
    img_path=img,
    actions=['age', 'gender'],
    )
print(f"Time: {time.time() - start:.5f}s")

# Load models
print("=====Load Split Models=====")
start = time.time()
model = AgeGenderModel()
print(f"Time: {time.time() - start:.5f}s")

# Batch Predict
start = time.time()
genders, ages = model.predicts(imgs)
print("=====Batch Predict=====")
print(genders, ages)
print(f"Time: {time.time() - start:.5f}s")

# Single Predict
start = time.time()
gender, age = model.predict(img)
print("=====Single Predict=====")
print(gender, age)
print(f"Time: {time.time() - start:.5f}s")
```

`predicts` is placed in age and gender clients to keep the `DeepFace.analyze` function logic.

Thank you for taking time to go through this feedback. :)


h-alice:
Bump
Really needs this feature.

serengil:
I don't support having another predicts function. Instead, you can add that logic under predict.

1- predict accepts both single image and list of images as

```python
img: Union[np.ndarray, List[np.ndarray]]
```

2- in predict function, you can check the type of img, and redirect it to your logic if it is list as

```python
if isinstance(img, np.ndarray):
   # put old predict logic here
elif isinstance(img, np.ndarray):
   # put your batch processing logic here
```

3- this new logic is worth to have its own unit tests. possibly, you can add some unit tests [here](https://github.com/serengil/deepface/blob/master/tests/test_analyze.py).

deepface/models/demography/Age.py
```diff
@@ -37,11 +41,64 @@ def __init__(self):
...
+        Args:
+            img: Single image as np.ndarray (224, 224, 3) or
+                List of images as List[np.ndarray] or
+                Batch of images as np.ndarray (n, 224, 224, 3)
+        Returns:
+            np.ndarray (n,)
+        """
+        # Convert to numpy array if input is list
+        if isinstance(img, list):
+            imgs = np.array(img)
```
serengil, line 56: cast img to img itself please, in that way we will not need else condition

deepface/models/demography/Age.py
```diff
@@ -37,11 +41,64 @@ def __init__(self):
...
+        else:
+            imgs = img
+
+        # Remove batch dimension if exists
+        imgs = imgs.squeeze()
```
serengil, line 61: you squeezed imgs and then expand its dimension, instead can we decide the input is single or many images when predict called?

```python
if isinstance(img, list):
   # then it is many images
elif len(img.shape) == 3:
   # then it is single image
elif len(img.shape) == 4 and img.shape[0] == 1:
   # then it is single image
elif len(img.shape) == 4 and img.shape[0] > 1:
   # then it is many images
else:
   raise error


```

serengil:
Actions failed because of linting - [link](https://github.com/serengil/deepface/actions/runs/12556163507/job/35011109078?pr=1396)

```
************* Module deepface.models.demography.Age
pylint: Command line or configuration file:1: UserWarning: 'Exception' is not a proper value for the 'overgeneral-exceptions' option. Use fully qualified name (maybe 'builtins.Exception' ?) instead. This will cease to be checked at runtime in 3.1.0.
deepface/models/demography/Age.py:70:0: C0303: Trailing whitespace (trailing-whitespace)
************* Module deepface.models.demography.Emotion
deepface/models/demography/Emotion.py:88:0: C0303: Trailing whitespace (trailing-whitespace)
```

deepface/models/demography/Age.py
```diff
@@ -37,11 +41,64 @@ def __init__(self):
...
+        
+        # Calculate apparent ages
+        apparent_ages = np.array(
+            [find_apparent_age(age_prediction) for age_prediction in age_predictions]
+        )
+
+        return apparent_ages
+
+
+    def predicts(self, imgs: List[np.ndarray]) -> np.ndarray:
```
serengil, line 79: predicts not being used anymore

deepface/models/demography/Gender.py
```diff
@@ -37,10 +41,54 @@ def __init__(self):
...
+            # Single image - add batch dimension
+            imgs = np.expand_dims(imgs, axis=0)
+
+        # Batch prediction
+        predictions = self.model.predict_on_batch(imgs)
+
+        return predictions
+
+
+    def predicts(self, imgs: List[np.ndarray]) -> np.ndarray:
```
serengil, line 74: predicts not in use

deepface/models/demography/Age.py
```diff
@@ -37,11 +41,64 @@ def __init__(self):
...
+        # Remove batch dimension if exists
+        imgs = imgs.squeeze()
+
+        # Check input dimension
+        if len(imgs.shape) == 3:
+            # Single image - add batch dimension
+            imgs = np.expand_dims(imgs, axis=0)
+
+        # Batch prediction
+        age_predictions = self.model.predict_on_batch(imgs)
```
serengil, line 69: `model.predict` causes memory issue when it is called in a for loop, that is why we call it as `self.model.predict(img, verbose=0)[0, :]`

in your design, if this is called in a for loop, still it will cause memory problem.

IMO, if it is single image, we should call `self.model.predict(img, verbose=0)[0, :]`, it is many faces then call `self.model.predict_on_batch`

deepface/models/demography/Age.py
NatLee: Sorry, forget to remove it.

tests/test_analyze.py
```diff
@@ -135,3 +135,15 @@ def test_analyze_for_different_detectors():
                     assert result["gender"]["Man"] > result["gender"]["Woman"]
                 else:
                     assert result["gender"]["Man"] < result["gender"]["Woman"]
+
+def test_analyze_for_multiple_faces():
+    img = "dataset/img4.jpg"
+    # Copy and combine the same image to create multiple faces
+    img = cv2.imread(img)
+    img = cv2.hconcat([img, img])
+    demography_objs = DeepFace.analyze(img, silent=True)
```
serengil, line 144: you should check the length of demography_objs, it should be 2 because 2 images fed (i hope only one face detected in img4.jpg)

deepface/modules/demography.py
```diff
@@ -130,83 +130,107 @@ def analyze(
         anti_spoofing=anti_spoofing,
     )
 
+    # Anti-spoofing check
+    if anti_spoofing:
```
serengil, line 134: can be written cleaner as

```python
if anti_spoofing and any(img_obj.get("is_real", True) is False for img_obj in img_objs):
    raise ValueError("Spoof detected in the given image.")
```

deepface/modules/demography.py
```diff
@@ -130,83 +130,107 @@ def analyze(
...
+        return []
 
-            elif action == "race":
-                race_predictions = modeling.build_model(
-                    task="facial_attribute", model_name="Race"
-                ).predict(img_content)
-                sum_of_predictions = race_predictions.sum()
+    # Convert the list of valid faces to a numpy array
+    faces_array = np.array(valid_faces)
+    resp_objects = [{} for _ in range(len(valid_faces))]
```
serengil, line 166: please do not create resp_objects in advance, instead create resp_object in the loop and append it to resp_objects similar to previous design

deepface/models/demography/Age.py
```diff
@@ -37,11 +41,64 @@ def __init__(self):
...
+        if isinstance(img, list):
+            imgs = np.array(img)
+        else:
+            imgs = img
+
+        # Remove batch dimension if exists
+        imgs = imgs.squeeze()
+
+        # Check input dimension
+        if len(imgs.shape) == 3:
```
serengil, line 64: we are doing same preprocessing to img in age, gender, emotin and race modules.

instead, cannot we do this in upper module - demography.py?

deepface/models/demography/Age.py
serengil: hey, even though this is sorted in newer tf versions, many users using old tf versions raise tickets about this problem. so, we should consider the people using older tf version. that is why, i recommend to use `self.model(img, training=False).numpy()[0, :]` for single images, and `self.model.predict_on_batch` for batches.

deepface/modules/demography.py
```diff
@@ -130,83 +130,109 @@ def analyze(
...
 
-        resp_objects.append(obj)
+    # For each action, predict the corresponding attribute
+    pbar = tqdm(
+        range(0, len(actions)),
+        desc="Finding actions",
+        disable=silent if len(actions) > 1 else True,
+    )
+    
+    for index in pbar:
```
serengil, line 177: please create a resp_object in the for loop as

```python
for index in pbar:
   resp_object = {}
```

deepface/modules/demography.py
```diff
@@ -130,83 +130,109 @@ def analyze(
...
+                
+                resp_objects[idx]["dominant_emotion"] = Emotion.labels[np.argmax(predictions)]
+
+        elif action == "age":
+            # Build the age model
+            model = modeling.build_model(task="facial_attribute", model_name="Age")
+            age_predictions = model.predict(faces_array)
+            
+            for idx, age in enumerate(age_predictions):
+                resp_objects[idx]["age"] = int(age)
```
serengil, line 202: set resp_object instead of resp_objects as

```python
resp_object["age"] = int(age)
```

do this for each attribute please

deepface/modules/demography.py
serengil: i really don't like to set resp_objects with idx, recommend something in the code

deepface/modules/demography.py
```diff
@@ -130,83 +130,109 @@ def analyze(
...
+                resp_objects[idx]["race"] = {}
+                
+                for i, race_label in enumerate(Race.labels):
+                    race_prediction = 100 * predictions[i] / sum_of_predictions
+                    resp_objects[idx]["race"][race_label] = race_prediction
+                
+                resp_objects[idx]["dominant_race"] = Race.labels[np.argmax(predictions)]
+
+    # Add the face region and confidence to the response objects
+    for idx, resp_obj in enumerate(resp_objects):
```
serengil, line 234: similar, region and confidence should be set in resp_object, and resp_object should be appended to resp_objects

previous design for setting resp_object is better

deepface/modules/demography.py
NatLee: i simplified it with `zip` and `update` in https://github.com/serengil/deepface/commit/52a38ba21a9b7edb331e2a6f25e68115ca1a663c
it looks more clean now :)

deepface/modules/demography.py
```diff
@@ -130,83 +130,109 @@ def analyze(
...
-                obj["race"] = {}
-                for i, race_label in enumerate(Race.labels):
-                    race_prediction = 100 * race_predictions[i] / sum_of_predictions
-                    obj["race"][race_label] = race_prediction
+    # Convert the list of valid faces to a numpy array
+    faces_array = np.array(valid_faces)
 
-                obj["dominant_race"] = Race.labels[np.argmax(race_predictions)]
+    # Create placeholder response objects for each face
+    for _ in range(len(valid_faces)):
```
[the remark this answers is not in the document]
NatLee: resolve in https://github.com/serengil/deepface/pull/1396#discussion_r1904679610

deepface/modules/demography.py
NatLee: resolved in https://github.com/serengil/deepface/pull/1396#discussion_r1904679610

deepface/modules/demography.py
NatLee: are you referring to whether the attribute mentioned refers to the percentage of `emotion` or `gender`?

deepface/modules/demography.py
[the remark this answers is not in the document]
serengil: seems this is not resovled.

before for loop, you should create a resp_objects as `resp_objects = []`

in for loop, you should create resp_object as `resp_object = {}` and set its features.

in summary, instead `resp_objects[idx]["emotion"] = {}`, we should have something like `resp_object["emotion"] = {}`

deepface/modules/demography.py
serengil: i don't want you to use idx at all.

deepface/models/Demography.py
```diff
@@ -18,5 +18,31 @@ class Demography(ABC):
...
+        Preprocess single or batch of images, return as 4-D numpy array.
+        Args:
+            img: Single image as np.ndarray (224, 224, 3) or
+                List of images as List[np.ndarray] or
+                Batch of images as np.ndarray (n, 224, 224, 3)
+        Returns:
+            Four-dimensional numpy array (n, 224, 224, 3)
+        """
+        if isinstance(img, list): # Convert from list to image batch.
+            image_batch = np.array(img)
```
serengil, line 36: if you use it like this, you don't have to have if and else blocks.

```
img = alpha_embedding = np.asarray(img)
```

deepface/modules/demography.py
NatLee: i initial the `resp_objects` renamed as `results` with face region and confidence.
https://github.com/NatLee/deepface/blob/52a38ba21a9b7edb331e2a6f25e68115ca1a663c/deepface/modules/demography.py#L155-L157

and use `update` function to write result of models like this
https://github.com/NatLee/deepface/blob/52a38ba21a9b7edb331e2a6f25e68115ca1a663c/deepface/modules/demography.py#L180-L193

the newest version didn't use any `idx` in the code.

NatLee: ahh the newest version is in the other branch which hasnt been merged yet

deepface/models/Demography.py
NatLee: i think @h-alice 's concern is likely about the `img` variable being compromised. 🤔

h-alice: @serengil  We appreciate your advice.
We will implement your recommended approach later.

deepface/models/demography/Age.py
h-alice: Hi! 👋
Please take a look at our prediction function, which uses the legacy single prediction method you suggested, and also provides batch prediction if a batch of images is provided.

The prediction method also correctly identifies the input image channels and forwards them to the model.

Please let us know if there’s anything else we can improve. Any advice you have is greatly appreciated.

https://github.com/serengil/deepface/blob/8883b212b2a88787ab26e4015c0fac9e2aa71439/deepface/models/Demography.py#L24

deepface/models/demography/Age.py
h-alice: We resolved the issue in our current commit, please take a look.

deepface/modules/demography.py
NatLee: already merged in this PR

deepface/models/demography/Age.py
NatLee: We added a proecessing function here :)
https://github.com/serengil/deepface/pull/1396/files#diff-64877893bf53c8eb3c54861ceea26ad73f96377c92ad6b02f6fd5be96f5aeb40R50

deepface/modules/demography.py
```diff
@@ -9,7 +9,7 @@
 from deepface.modules import modeling, detection, preprocessing
 from deepface.models.demography import Gender, Race, Emotion
 
-
+# pylint: disable=trailing-whitespace
```
serengil, line 12: please do not disable this warning, if you have trailing whitespace issue then you can use a linter

deepface/modules/demography.py
```diff
@@ -130,83 +128,105 @@ def analyze(
...
-
-        resp_objects.append(obj)
-
-    return resp_objects
+            return None
+        img_content = img_content[:, :, ::-1]  # BGR to RGB
+        return preprocessing.resize_image(img=img_content, target_size=(224, 224))
+
+    # Filter out empty faces
+    face_data = [(preprocess_face(img_obj), img_obj["facial_area"], img_obj["confidence"])
```
serengil, line 145: i tried to explain my expectation in my previous comments but this behaviour is still missing:

- please build a for loop for img_objs
- in each iteraration, create a dict `resp_object` and set its demography results, region and face_confidence. then append it to `resp_objects` list.
- do not create face_data, valid_faces, face_regions, face_confidences in advance. everything should be done in the iteration.

NatLee: Thank you for the advice. But I'm still curious to understand why this approach is considered better. Could you please provide some insights?

serengil: I have to handle the bugs. So, I have to understand the source code. So, that is cleaner for me.

In the past, I approved many PRs but they caused bugs. I spent much more time to understand what has been done.

deepface/modules/demography.py
NatLee: Okay! Solved.

deepface/modules/demography.py
```diff
@@ -175,6 +175,7 @@ def analyze(
                     task="facial_attribute", model_name="Age"
                 ).predict(img_content)
                 # int cast is for exception - object of type 'float32' is not JSON serializable
+                print(apparent_age.shape)
```
serengil, line 178: print forgotten here

deepface/modules/demography.py
NatLee: Can we create a dedicated function specifically for batch processing, named something like `batch_analyze`, and place it in the same file as `analyze`?

This way, it wouldn’t interfere with the existing `analyze` function and would also support batch input, similar to how Keras implements `predict_batch`.

serengil: Prefer to have single analyze

NatLee: Thank you for quickly response.

Adding a variable named `batch` or other to the analyze function, with a default value set to False, seems like a practical solution.

By doing this, the new functionality can only be accessed when batch is explicitly set to True. This approach ensures backward compatibility, as existing users who rely on the current behavior of the function will remain unaffected.

Could you provide your perspective on this matter?

serengil: i disagree, we should decide to perform single or batch according to the incoming input

NatLee: Is it good on executing our function only after determining that the input for `analysis` is plural? Like list of string or multiple images.

h-alice: @serengil  Sure! We’re currently working on our purposed `analysis` function to make it more flexible and yet keep the original maintainability. 
It might take a bit of time, but we’ll let you know when we're done.

serengil: please let us to discuss first before pushing

h-alice: @serengil Thank you so much for accepting our PR! 
We would greatly appreciate your assistance in merging it to the master branch. 

For the `analysis` function we are currently working on, we will keep you updated as we make more progress. 
Once it’s completed, we’ll create a new PR for it.

Thank you once again for your review of our code.

serengil: I will make a final review to this today, and merge it after - if I don't have any concerns.

For the new PR, I am not sure about that feature. Do you mind to open a ticket for this and discuss the topic under it? Otherwise, if i am not satisfied with the feature, I would close it and your time will be wasted.

tests/test_analyze.py
```diff
@@ -135,3 +135,16 @@ def test_analyze_for_different_detectors():
                     assert result["gender"]["Man"] > result["gender"]["Woman"]
                 else:
                     assert result["gender"]["Man"] < result["gender"]["Woman"]
+
+def test_analyze_for_multiple_faces():
+    img = "dataset/img4.jpg"
+    # Copy and combine the same image to create multiple faces
+    img = cv2.imread(img)
+    img = cv2.hconcat([img, img])
```
serengil, line 143: hconcat makes a single image

```
input image shape before hconcat is (1728, 2500, 3)
input image shape after hconcat is (1728, 5000, 3)
```

to have a numpy array with (2, 1728, 2500, 3) shape, you should do something like:

```
img = np.stack((img, img), axis=0)
```

deepface/models/Demography.py
```diff
@@ -18,5 +18,54 @@ class Demography(ABC):
...
+        Args:
+            img: Single image as np.ndarray (224, 224, 3) or
+                 List of images as List[np.ndarray] or
+                 Batch of images as np.ndarray (n, 224, 224, 3)
+        Returns:
+            Four-dimensional numpy array (n, 224, 224, 3)
+        """
+        image_batch = np.array(img)
+        # Remove batch dimension in advance if exists
+        image_batch = image_batch.squeeze()
```
serengil, line 66: i initially confused about why why squeeze first and expand dimensions second

would you please add a comment here something like:

we did perform squeeze and expand dimensions sequentially to have same behaviour for (224, 224, 3), (1, 224, 224, 3) and (n, 224, 224, 3)

tests/test_analyze.py
serengil: Also please check that img is now having (2, x, x, x) shape

serengil: Finally, unit tests failed for that input. The case you tested did not test what you did. It is still single image.

deepface/models/Demography.py
h-alice: We took a look at the processing flow and discovered that the `squeeze` operation is unnecessary. Every single image input would have an expanded batch dimension of `(1, 224, 224, 3)`, so there’s no need to handle inputs with this dimension.

The redundant `squeeze` process has been removed.

deepface/modules/demography.py
NatLee: Understood, let's complete this PR first and then open a separate issue for further discussion.

tests/test_analyze.py
h-alice: Hi @serengil 👋
We have implemented batched images support in `DeepFace::analysis`, and the test cases have been modified as per your request. Please help us check if this matches your requirements.

Due to the complexity of designing a more efficient flow for `analysis`, we will prioritize extending the functionality of models that can accept batched images for now.

We will discuss enhancing the performance of the `analysis` function in separate threads or through pull requests. We would invite you to participate in these discussions once we are ready.

Please help us merge this PR if all the requirements are met.

serengil:
I am not available to review it until early Feb.

serengil:
Tests failed

NatLee:
The test seems passed. If you're okay, it's ready to merge.

deepface/DeepFace.py
```diff
@@ -253,6 +256,29 @@ def analyze(
             - 'middle eastern': Confidence score for Middle Eastern ethnicity.
             - 'white': Confidence score for White ethnicity.
     """
+
+    if isinstance(img_path, np.ndarray) and len(img_path.shape) == 4:
```
serengil, line 260: this control should not be done in DeepFace.py

as you can see, we stored no logic in this file.

NatLee: Alright, I've moved the control logic into the analyze function.

serengil:
LGTM

h-alice:
Thank you so much @serengil !

NatLee:
@serengil Many thanks for your reviews!

Issue #1406, opened 2024-12-19
[FEATURE]: align_img_wrt_eyes performance improvement


sinopec:
### Description

When I use extract_faces  with align set to be True by default, I found it really slow  for the whole process,
It takes about 1.45 secs while the input img is  2100x1275 with 2 people in  and the detector is yolov8.

I track the whole process and found :
yolov8  cost  0.03 secs
align_img_wrt_eyes  cost about 0.7*2 secs —— two people
What should I do if I want speed up 

use smaller  input image will sure speed up, but  can I do more?
If I turned off align flag ,what will happen?

huulockt:
@serengil Could you please review it and let me know what you think?

serengil:
Would you please run your code for the items in the following issues?

- Issue 1: https://github.com/serengil/deepface/issues/1244
- Issue 2: https://github.com/serengil/retinaface/issues/98

huulockt:
@serengil  tested with **centerface** because there were some problems building **retinaface**. Here are the results (represented as `<face, confidence, (x,y,w,h), right eye coords, left eye coords>`):
### Issue 1:
- ![test1_0](https://github.com/user-attachments/assets/0818466f-afd9-4194-9ed3-970b476e64ae), (1740 301 149 147), 0.907188892364502, (1800, 346), (1849, 380)
- ![test1_1](https://github.com/user-attachments/assets/ecd31e92-f258-4c34-97b3-00f76ab7dabf), (1512 137 152 165), 0.8301119804382324, (1577, 188), (1630, 222)
### Issue 2:
- ![test2_0](https://github.com/user-attachments/assets/8262496c-db1e-448a-a887-c63f34ff6592), (423 240 55 70), 0.9126578569412231, (437, 267), (464, 266)
- ![test2_1](https://github.com/user-attachments/assets/28450543-ec95-44f1-be2f-eaf0552965a3), (187 232 59 70), 0.8854385018348694, (206, 257), (233, 258)
- ![test2_2](https://github.com/user-attachments/assets/3374f9ec-f39d-409b-9011-726cb9eb0fa2), (1 88 47 79), 0.8846514225006104, (7, 113), (36, 113)
- ![test2_3](https://github.com/user-attachments/assets/84405723-0d49-4645-b647-6ba84789ba7d), (523 70 56 76), 0.8671445250511169, (540, 100), (567, 100)

serengil:
@huulockt very good then. would you please create a PR? Your name should appear in the contributors.

[...]

Issue #1244, opened 2024-06-02
[BUG]: Alignment process may not always work as expected


PhilippeRo:
### Before You Report a Bug, Please Confirm You Have Done The Following...

- [X] I have updated to the latest version of the packages.
- [X] I have searched for both [existing issues](https://github.com/serengil/deepface/issues) and [closed issues](https://github.com/serengil/deepface/issues?q=is%3Aissue+is%3Aclosed) and found none that matched my issue.

### DeepFace's version

0.0.91

### Python version

3.11

### Operating System

Fedora 40

### Dependencies

absl-py==2.1.0
aiohttp==3.9.1
aiosignal==1.3.1
aiostream==0.5.2
annotated-types==0.6.0
anyio==3.7.1
appdirs==1.4.4
argcomplete==3.3.0
astunparse==1.6.3
attrs==23.1.0
Authlib==1.3.0
Babel==2.15.0
backoff==2.2.1
bcrypt==4.0.1
Beaker==1.12.1
beautifulsoup4==4.12.3
black==23.11.0
blinker==1.7.0
blivet==3.10.0
blivet-gui==2.5.0
bodhi-client==8.1.0
boto3==1.34.99
botocore==1.34.99
bottle==0.12.25
Brlapi==0.8.5
Brotli==1.1.0
cachetools==5.3.2
CCColUtils==1.5
certifi==2023.11.17
cffi==1.16.0
chardet==5.2.0
charset-normalizer==3.3.2
click==8.1.7
colorama==0.4.6
construct==2.10.70
contourpy==1.2.0
cryptography==41.0.7
cupshelpers==1.0
cycler==0.11.0
dasbus==1.7
dataclasses-json==0.6.3
dbus-python==1.3.2
decorator==5.1.1
deepface==0.0.90
defusedxml==0.7.1
Deprecated==1.2.14
diskcache==5.6.3
distro==1.9.0
dm-tree==0.1.8
dnf==4.19.2
docutils==0.20.1
dropbox==11.36.2
duplicity==2.2.3
fastapi==0.104.1
fasteners==0.19
fedora-third-party==0.10
fedpkg==1.44
file-magic==0.4.0
filelock==3.13.1
fire==0.5.0
Flask==3.0.2
flatbuffers==23.5.26
fonttools==4.50.0
fros==1.1
frozenlist==1.4.0
fs==2.4.16
fsspec==2023.12.1
gast==0.5.4
gbinder-python==1.1.2
gdown==5.1.0
[...]

PR #1409, opened 2024-12-23
Improve the face alignment performance in detect_faces()

huulockt:
## Tickets
#1244
#1406

### What has been done

With this PR, the detect_faces() logic has been modified when alignment is enabled:
- The original image is passed directly to the detect function, so the detection results are not affected by the alignment flag.
- Only the face region is aligned, rather than the entire image, which improves alignment speed.

Sorry I didn't add a unit test case for #1244 as I promised. I think this bug can only be detected visually, since it’s hard to test automatically. But if an automated test is needed, I’d suggest using template matching algorithms.
## How to test

```shell
make lint && make test
```

### DIFF
--- a/deepface/modules/detection.py
+++ b/deepface/modules/detection.py
@@ -301,10 +301,15 @@ def expand_and_align_face(
     detected_face = img[int(y) : int(y + h), int(x) : int(x + w)]
     # align original image, then find projection of detected face area after alignment
     if align is True:  # and left_eye is not None and right_eye is not None:
-        aligned_img, angle = align_img_wrt_eyes(img=img, left_eye=left_eye, right_eye=right_eye)
+        processed_face, new_x, new_y = alignment_preprocess(img=img, facial_area=(x,y,w,h))
+        aligned_img, angle = align_img_wrt_eyes(
+            img=processed_face, left_eye=left_eye, right_eye=right_eye
+        )
 
         rotated_x1, rotated_y1, rotated_x2, rotated_y2 = project_facial_area(
-            facial_area=(x, y, x + w, y + h), angle=angle, size=(img.shape[0], img.shape[1])
+            facial_area=(new_x, new_y, new_x + w, new_y + h),
+            angle=angle,
+            size=(processed_face.shape[0], processed_face.shape[1])
         )
         detected_face = aligned_img[
             int(rotated_y1) : int(rotated_y2), int(rotated_x1) : int(rotated_x2)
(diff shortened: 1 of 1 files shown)


serengil:
When I perform current design and your change of detection module for [this image](https://github.com/serengil/deepface/blob/master/tests/dataset/selfie-many-people.jpg), I am getting these results:

```shell
current design:

opencv deteceted 5 faces
ssd deteceted 3 faces
dlib deteceted 4 faces
mtcnn deteceted 6 faces
retinaface deteceted 7 faces
yunet deteceted 4 faces
yolov8 deteceted 8 faces
centerface deteceted 6 faces

your change:
opencv deteceted 5 faces
ssd deteceted 4 faces
dlib deteceted 4 faces
mtcnn deteceted 5 faces
retinaface deteceted 6 faces
yunet deteceted 4 faces
yolov8 deteceted 7 faces
centerface deteceted 7 faces
```

You can also find the detection results here: 
Screenshot 2024-12-24 at 11 01 49

As you can see, your change causes not to find faces close to image boundaries for most of detectors.

I am sharing my test code here:

```python
from deepface import DeepFace

img_path = "dataset/selfie-many-people.jpg"

detector_backends = [
    "opencv",
    "ssd",
    "dlib",
    "mtcnn",
    "retinaface",
    "yunet",
    "yolov8",
    "centerface",
]

for detector_backend in detector_backends:
    face_objs = DeepFace.extract_faces(
        img_path=img_path,
        detector_backend=detector_backend,
        # expand_percentage=0,
    )
    print(f"{detector_backend} deteceted {len(face_objs)} faces")
    fig = plt.figure(figsize=(10, 10))
    for face_obj in face_objs:
        face = face_obj["face"]
        plt.imshow(face)
        plt.axis("off")
        plt.show()
```

TLDR: current design can detect faces close to boundaries but your change cannot.

huulockt:
Thanks for the feedback! I'm busy with Christmas right now, but I’ll check it carefully soon. For now, here are my thoughts:

- Each model is pretrained on its own dataset, so there are always some constraints on what it can detect.
- Lightweight detectors like SSD, YuNet, and CenterFace can struggle with large images. In my design, by not adding border before detection helps, these models tend to perform better.
- Detection results mainly depend on the threshold. If we want to make it easier for users,  we could run a proper benchmark to suggest optimal thresholds. Otherwise, users would need to find an appropriate threshold themselves for their specific models  (as I'm doing).

P/S: Merry Christmas! Hope you enjoy the holiday season. 🎄

serengil:
Of course, take your time. I hope you understand my concern. When increasing its time consumption performance, I don't want to decrease its detection performance. Enhancement should offer same accuracy performance or more. Here are my comments:

- Each model is pretrained on its own dataset, so there are always some constraints on what it can detect -> This is independent from model because with current design retinaface and mtcnn can detect more faces. So, your change caused this.
- Lightweight detectors like SSD, YuNet, and CenterFace can struggle with large images. In my design, by not adding border before detection helps, these models tend to perform better -> right, ssd and centerface outperform than existing design.
- Detection results mainly depend on the threshold -> again, we used same threshold for retinaface and mtcnn, but new design is missing some faces. So, that must be related to the detection logic you proposed.

huulockt:
Actually, the detection results in my design are the same as the current design when the `align` flag is turned off. Here’s the first solution I came up with: We can keep the current border-adding step before detection, but combine it with my proposal to only apply alignment to the face region. I implemented this in the last commit. However, when I tested it myself, I couldn’t figure out why mtcnn still returns 5 faces in both my design and the current design. Could you please run the test code again with mtcnn?

Moreover, the above solution doesn’t fully preserve the detection improvements observed with models like ssd, centerface, and yunet(*). I propose adding these models in a skip-border-addition list, and the code would look like this:
`if align is True and model_name not in skip_list:`

Additionally, for further clarification, in my design, yolov8 can detect all 7 faces, but faces near the border return outer-eye coordinates as (0,0), which affects alignment. The current design didn't have this problem because of the border, so please let me know if you want to add yolov8 to the skip-list in the list too.

What do you think of these ideas?

(*) For yunet, with a threshold of 0.8, the current design detects 4 faces, while my design detects 6. Perhaps we could consider lowering the threshold for better results.

serengil:
So, we are not adding borders to image. Why should we merge this PR then? I cannot see any reasonable improvements.

huulockt:
The main improvement in my design lies in how the alignment input is adjusted, regardless of whether borders are added or not. Currently, the entire image is used as the input for alignment, and this process is repeated `n` times—once for each face detected.

In my design, only the facial area is used as the input for each alignment operation. Since the entire image is significantly larger than the facial area, this optimization saves a considerable amount of processing time—especially when multiple faces are present in the image.

I hope this explanation clarifies the benefits of the proposed changes.

serengil:

serengil:
I just updated the function name and variable names to make them more meaningful, and added a comment about this change: https://github.com/serengil/deepface/commit/3d914f13370cd26cdedc454e3ff7388fbe6fffde

huulockt:
Of course, I did struggle with naming the function and variables. I had considered something like `expand_facial_area()`, but that would likely mean sharing it with the existing expansion step above. For the variables, `new_x` and `new_y` represent coordinates adjusted relative to the expanded region. I thought about names like `expanded_facial_x` and `expanded_facial_y`, but they felt too long and potentially confusing.

serengil:
current variable and function names are better - i think

https://github.com/serengil/deepface/blob/master/deepface/modules/detection.py

huulockt:
It looks great!

serengil:
appreciate again

PR #1412, opened 2024-12-28
Improved flexibility for recurring operations

freakmaxi:
## Tickets

No ticket

### What has been done

Just improved the flexibility for the recurring actions on the same image with same faces. So, when the faces are extracted once, they can be used on all other operations which help to accelerate the operation and make significant performance improvement for the series of operation

## How to test

```shell
make lint && make test
```

### DIFF
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -262,6 +262,83 @@ def analyze(
     )
 
 
+def analyze_from_image_objects(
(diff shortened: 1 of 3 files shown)


serengil:
Hello, thank you for your contribution but TBH I don't want to add new functions into interface.

You should raise a ticket and discuss this feature before starting development as recommended in the read me.

Issue #1413, opened 2024-12-29
[BUG]: I rotated the image by 90 degrees clockwise just to test if Deepface properly aligns that


jogendra-india:
### Before You Report a Bug, Please Confirm You Have Done The Following...

- [X] I have updated to the latest version of the packages.
- [X] I have searched for both [existing issues](https://github.com/serengil/deepface/issues) and [closed issues](https://github.com/serengil/deepface/issues?q=is%3Aissue+is%3Aclosed) and found none that matched my issue.

### DeepFace's version

v0.0.94 (directly cloned the master branch)

### Python version

3.9

### Operating System

Ubuntu

### Dependencies

absl-py==2.1.0
amqp==5.3.1
asgiref==3.8.1
asttokens==3.0.0
astunparse==1.6.3
async-timeout==5.0.1
backcall==0.2.0
beautifulsoup4==4.12.3
billiard==4.2.1
black==24.10.0
blinker==1.9.0
celery==5.4.0
certifi==2024.12.14
cfgv==3.4.0
charset-normalizer==3.4.1
click==8.1.8
click-didyoumean==0.3.1
click-plugins==1.1.1
click-repl==0.3.0
contourpy==1.3.0
cycler==0.12.1
decorator==5.1.1
deepface @ git+https://github.com/serengil/deepface.git@ca6afd98efb87dc61fb022ddd2892da23dc75f68
distlib==0.3.9
Django==4.2
django-celery-results==2.5.1
django-cors-headers==4.6.0
django-extensions==3.2.3
django-filter==24.3
djangorestframework==3.15.2
executing==2.1.0
filelock==3.16.1
fire==0.7.0
Flask==3.1.0
Flask-Cors==5.0.0
flatbuffers==24.12.23
fonttools==4.55.3
gast==0.6.0
gdown==5.2.0
google-pasta==0.2.0
greenlet==3.1.1
grpcio==1.68.1
gunicorn==23.0.0
h5py==3.12.1
identify==2.6.3
idna==3.10
importlib_metadata==8.5.0
importlib_resources==6.4.5
ipython==8.12.3
itsdangerous==2.2.0
jedi==0.19.2
Jinja2==3.1.5
keras==3.7.0
kiwisolver==1.4.7
kombu==5.4.2
libclang==18.1.1
Markdown==3.7
markdown-it-py==3.0.0
MarkupSafe==3.0.2
matplotlib==3.9.4
matplotlib-inline==0.1.7
mdurl==0.1.2
[...]

PR #1414, opened 2024-12-30
Refactor face extraction and alignment logic; enhance bounding box transformation and facial landmark adjustments

jogendra-india:
## Tickets

https://github.com/serengil/deepface/issues/1413

### What has been done
In this PR, I have modified the code so that bounding box around image also gets rotated on image rotation during alignment. 

Here are some result:
Original Image I used (left rotated):
image_rotated

Here is the earlier aligned image extracted (plotted aligned_sub_img from extract_face() function that is return of align_img_wrt_eyes function)
Earlier aligned image

After the change in PR, this is how it looks now:
Final aligned image after change

Tested the performance on random images to identify the same person using embeddings from the image below.
Original image

Image 1:
Image 1

Image 2:
Image 2

Image 3:
Image 3

Here are the results:

### Cosine Similarity Comparison Table

| Metric      | Image 1 (Original) | Image 1 (After Change) | Image 2 (Original) | Image 2 (After Change) | Image 3 (Original) | Image 3 (After Change) |
|-------------|---------------------|-------------------------|---------------------|-------------------------|---------------------|-------------------------|
| **Facenet** | 0.63                | 0.65                    | 0.78                | 0.75                    | 0.48                | 0.41                    |
| **ArcFace** | 0.45                | 0.39                    | 0.47                | 0.42                    | 0.29                | 0.19                    |
| **VGG-Face**| 0.40                | 0.39                    | 0.46                | 0.42                    | 0.64                | 0.73                    |
| **Facenet512** | 0.69              | 0.63                    | 0.66                | 0.58                    | 0.85                | 0.90                    |

### Observations:
1. **Image 1** shows a slight improvement in **Facenet** (0.63 → 0.65) but a slight drop in other metrics.
2. **Image 2** maintains overall performance but slightly decreases across all metrics after the change.
3. **Image 3** (left-rotated) demonstrates better alignment with **VGG-Face** (0.64 → 0.73) and **Facenet512** (0.85 → 0.90), indicating improved handling of rotations.

@serengil Not entirely sure if this has resulted in significant improvements. While the alignment issue appears to be resolved, there may still be other areas requiring adjustments. I had anticipated cosine similarity improvements across all models, but the results seem mixed.

I would appreciate it if you could thoroughly test this with your expertise and recommend any additional changes that might be necessary.

Thank you!

## How to test

```shell
make lint && make test
```

### DIFF
--- a/deepface/modules/detection.py
+++ b/deepface/modules/detection.py
@@ -286,6 +286,7 @@ def extract_face(
     mouth_left = facial_area.mouth_left
     mouth_right = facial_area.mouth_right
 
+    # Expand the facial area if needed
     if expand_percentage > 0:
         # Expand the facial region height and width by the provided percentage
         # ensuring that the expanded region stays within img.shape limits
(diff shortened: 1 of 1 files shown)


serengil:
Hmmm, the scale of these changes seems risky given the minimal or negative impact. It is not worth the trouble.

jogendra-india:
> Hmmm, the scale of these changes seems risky given the minimal or negative impact. It is not worth the trouble.

Agreed! Alternatively, we could limit this change to cases where the eyes are positioned vertically, such as in rotated images. However, this might be over-engineering. I'll revisit this later in the week if I come up with other adjustments. How does it look on your end with your test images?

serengil:
seems it passed the unit tests

PR #1420, opened 2025-01-06
Optimizations for deepface.recognition.find, Optimization and New Iterator Functionality in image_utils

PyWoody:
### What has been done

With this PR, I have made optimizations and opinionated changes to `deepface.modules.recognition` as well as `deepface.commons.image_utils`. The optimizations include better use of sets, delayed object creation, and ahead of time creation for set objects used in tight loops.

The modifications proposed in this PR are highly opinionated, so please feel free to request any modifications or to refuse it outright.

#### Modified
`list_images`

- Image and PIL file extensions are now stored in sets built ahead of walking the directory. This prevents new sets from needing to be created on each iteration.
- exact_path is not created unless the file ext is already in image_exts

#### New
`yield_images`

- New generator function
- Mimics the same functionality as `list_images` but yields the exact_path instead of building and returning the list
- Using an iterator will improve speed and reduce memory when created as a set in `deepface.modules.recognition`

&nbsp;

#### Modified

- storage_images is now built using the new `deepface.commons.image_utils.yield_images` iterator
- storage_images is created as a set instead of a list, as all operations on it were set operations
- pickled_images is now delayed until `if refresh_database` is True'sh to prevent unnecessary object creation
- pickled_images is now created as a set comprehension instead of a list comprehension, as all operations on it were set operations
- new_images and old_images can now be created without having to convert pickled_images and storage_images into sets
- df_cols is stored as a set. If column order will be necessary in the future, this should be restored to a list

## How to test

The test in `deepface.tests.test_find.test_filetype_for_find_bulk_embeddings` checks the new `deepface.commons.image_utils.yield_images` for functionality and equivalence to the existing `deepface.commons.image_utils.list_images` function.

```shell
make lint && make test
```

### DIFF
--- a/deepface/commons/image_utils.py
+++ b/deepface/commons/image_utils.py
@@ -1,7 +1,7 @@
 # built-in dependencies
 import os
 import io
-from typing import List, Union, Tuple
+from typing import Generator, List, Union, Tuple
 import hashlib
 import base64
 from pathlib import Path
@@ -33,6 +33,26 @@ def list_images(path: str) -> List[str]:
                     if img.format.lower() in pil_exts:
                         images.append(exact_path)
     return images
+
+
+def yield_images(path: str) -> Generator[str]:
+    """
+    List images in a given path
+    Args:
+        path (str): path's location
+    Yields:
+        image (str): image path
+    """
+    images = []
+    image_exts = {".jpg", ".jpeg", ".png"}
+    pil_exts = {"jpeg", "png"}
+    for r, _, f in os.walk(path):
+        for file in f:
+            if os.path.splitext(file)[1].lower() in image_exts:
+                exact_path = os.path.join(r, file)
+                with Image.open(exact_path) as img:  # lazy
+                    if img.format.lower() in pil_exts:
+                        yield exact_path
 
 
 def find_image_hash(file_path: str) -> str:
--- a/deepface/commons/image_utils.py
+++ b/deepface/commons/image_utils.py
@@ -35,7 +35,7 @@ def list_images(path: str) -> List[str]:
     return images
 
 
-def yield_images(path: str) -> Generator[str]:
+def yield_images(path: str) -> Generator[str, None, None]:
     """
     Yield images in a given path
     Args:
--- a/deepface/commons/image_utils.py
+++ b/deepface/commons/image_utils.py
@@ -43,7 +43,6 @@ def yield_images(path: str) -> Generator[str]:
     Yields:
         image (str): image path
     """
-    images = []
     image_exts = {".jpg", ".jpeg", ".png"}
     pil_exts = {"jpeg", "png"}
     for r, _, f in os.walk(path):
--- a/deepface/commons/image_utils.py
+++ b/deepface/commons/image_utils.py
@@ -23,18 +23,15 @@ def list_images(path: str) -> List[str]:
         images (list): list of exact image paths
     """
     images = []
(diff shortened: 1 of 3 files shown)


deepface/modules/recognition.py
```diff
@@ -157,18 +157,15 @@ def find(
...
                 f"{i}-th item does not have some required keys - {missing_keys}."
                 f"Consider to delete {datastore_path}"
             )
 
-    # embedded images
-    pickled_images = [representation["identity"] for representation in representations]
-
     # Get the list of images on storage
-    storage_images = image_utils.list_images(path=db_path)
+    storage_images = set(image_utils.yield_images(path=db_path))
```
serengil, line 168: yield_images is meaningful if the number of images in the target folder too much

can we check the number of files in the target folder, use list images if it is not too much, and use yield_images if it is too much?

tests/test_find.py
```diff
@@ -95,12 +95,23 @@ def test_filetype_for_find():
...
+
+    # Generator
+    gen_imgs = list(image_utils.yield_images("dataset"))
+
+    assert len(gen_imgs) > 0
+
+    # img47 is webp even though its extension is jpg
+    assert "dataset/img47.jpg" not in gen_imgs
+
+    assert gen_imgs == list_imgs
```
serengil, line 114: good to see they are identical

deepface/commons/image_utils.py
```diff
@@ -23,19 +23,35 @@ def list_images(path: str) -> List[str]:
         images (list): list of exact image paths
     """
     images = []
+    image_exts = {".jpg", ".jpeg", ".png"}
```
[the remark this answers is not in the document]
PyWoody: I have no argument against it. There will be a very slight performance hit as it'll have to go to the next namespace but it should be negligible.

Below is a simple, silly example of the penalty for global vs local lookups. On my system, global was always slower but only marginally.

```python3
import itertools
import time


IMAGE_EXTS = {'.jpg', '.jpeg', '.png'}
PIL_EXTS = {'jpeg', 'png'}


def local_test():
    image_exts = {'.jpg', '.jpeg', '.png'}
    pil_exts = {'jpeg', 'png'}
    exts = itertools.cycle(
        ['.jpg', '.fake', '.png', '.tmp', '.jpeg', '.placeholder']
    )
    for _ in range(100_000_000):
        psuedo_ext = next(exts)
        if psuedo_ext in image_exts:
            if psuedo_ext.lstrip('.') in pil_exts:
                yield None


def global_test():
    exts = itertools.cycle(
        ['.jpg', '.fake', '.png', '.tmp', '.jpeg', '.placeholder']
    )
    for _ in range(100_000_000):
        psuedo_ext = next(exts)
        if psuedo_ext in IMAGE_EXTS:
            if psuedo_ext.lstrip('.') in PIL_EXTS:
                yield None


start = time.perf_counter()
_ = set(local_test())
print(f'LOCAL: {time.perf_counter() - start}')

start = time.perf_counter()
_ = set(global_test())
print(f'GLOBAL: {time.perf_counter() - start}')

```

If that's acceptable, I'd be glad to modify the commit.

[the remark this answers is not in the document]
serengil: No, i mean moving line 26 and 27 to after imports in this file and get rid of line 46 and 47

deepface/modules/recognition.py
serengil: never mind this, use yield one always

PR #1423, opened 2025-01-09
load_image now accepts file objects that support being read

PyWoody:
### What has been done

With this PR, `deepface.commons.image_utils.load_image` now accepts file objects that support being read.

Objects that support being read but not seeked are handled by reading the data into a new `io.BytesIO` object which is automatically closed before returning the data or raising an exception.

### PIL Image

Below is an example of opening a `PIL.Image` in memory, rotating the image, writing the modified image to an in-memory `io.BytesIO` object, and then performing a `DeepFace.represent` calculation on the in-memory object.

Modifications of this ilk could be converting images to greyscale, resizing, cropping, etc., all done in memory while not having to constantly write to disk.

```python3
from deepface import DeepFace
from PIL import Image

representations = []
with Image.open(r'/path/to/image.png') as img:
    for rotation in range(4):
        pil_io = io.BytesIO()
        rotated_img = img.rotate(rotation * 90)
        rotated_img.save(pil_io, format='PNG')
        rp = DeepFace.represent(pil_io, enforce_detection=False)
        representations.append(rp)
        pil_io.close()
```

#### Streaming object

I know `deepface.commons.image_utils.load_image` already accepts a URL as a string but it can be more convenient to control how the request is made yourself than by the library, i.e., allowing redirects, controlling timeout, handling tokens with Sessions,  etc.

```python3
import requests

from deepface import DeepFace

r = requests.get('https://raw.githubusercontent.com/serengil/deepface/refs/heads/master/tests/dataset/img1.jpg', stream=True)
r.raise_for_status()
representations = DeepFace.represent(r.raw)
```

#### As a File Object

```python3
with open(r'/path/to/image.png', 'rb') as f:
    representation = DeepFace.represent(f)
```

or

```python3
representation = DeepFace.represent(open(r'/path/to/image.png', 'rb'))
```

This functionality, while already provided by passing the filepath as a string, closer mirrors the functionality of other Python libraries, such as how `json.load`[0], `pickle.load`[1] and `PIL.Image.open`[2] operate.

#### From a ZipArchive

```python3
from zipfile import ZipFile

with ZipFile('some_archive.zip') as z:
    with z.open(r'/path/to/image.png') as f:
        representation = DeepFace.represent(f)
```

[0] https://docs.python.org/3/library/json.html#json.load
[1] https://docs.python.org/3/library/pickle.html#pickle.load
[2] https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.open

## How to test

A new test at `deepface.tests.test_represent.test_standard_represent_with_io_object` has been created to ensure representations loaded with the new `deepface.commons.image_utils.load_image_from_io_object` are identical to the default implementation. The new test also ensures objects that do not support `seek`, e.g., streaming objects, are also properly handled and supported.

```shell
make lint && make test
```

### DIFF
--- a/tests/test_represent.py
+++ b/tests/test_represent.py
@@ -1,4 +1,5 @@
 # built-in dependencies
+import io
 import cv2
 
 # project dependencies
@@ -16,6 +17,21 @@ def test_standard_represent():
         logger.debug(f"Function returned {len(embedding)} dimensional vector")
         assert len(embedding) == 4096
     logger.info("✅ test standard represent function done")
+
+
+def test_standard_represent_with_io_object():
+    img_path = "dataset/img1.jpg"
+    defualt_embedding_objs = DeepFace.represent(img_path)
+    io_embedding_objs = DeepFace.represent(open(img_path, 'rb'))
+    assert defualt_embedding_objs == io_embedding_objs
+
+    # Confirm non-seekable io objects are handled properly
+    io_obj = io.BytesIO(open(img_path, 'rb').read())
+    io_obj.seek = None
+    no_seek_io_embedding_objs = DeepFace.represent(io_obj)
+    assert defualt_embedding_objs == no_seek_io_embedding_objs
+
+    logger.info("✅ test standard represent with io object function done")
 
 
 def test_represent_for_skipped_detector_backend_with_image_path():
--- a/tests/test_represent.py
+++ b/tests/test_represent.py
@@ -1,6 +1,7 @@
 # built-in dependencies
 import io
 import cv2
+import pytest
 
 # project dependencies
 from deepface import DeepFace
@@ -30,6 +31,10 @@ def test_standard_represent_with_io_object():
     io_obj.seek = None
     no_seek_io_embedding_objs = DeepFace.represent(io_obj)
     assert default_embedding_objs == no_seek_io_embedding_objs
+
+    # Confirm non-image io objects raise exceptions
+    with pytest.raises(ValueError, match='Failed to decode image'):
+        DeepFace.represent(io.BytesIO(open(__file__, 'rb').read()))
 
     logger.info("✅ test standard represent with io object function done")
 
--- a/tests/test_represent.py
+++ b/tests/test_represent.py
@@ -21,15 +21,15 @@ def test_standard_represent():
 
 def test_standard_represent_with_io_object():
     img_path = "dataset/img1.jpg"
-    defualt_embedding_objs = DeepFace.represent(img_path)
+    default_embedding_objs = DeepFace.represent(img_path)
     io_embedding_objs = DeepFace.represent(open(img_path, 'rb'))
-    assert defualt_embedding_objs == io_embedding_objs
+    assert default_embedding_objs == io_embedding_objs
 
     # Confirm non-seekable io objects are handled properly
     io_obj = io.BytesIO(open(img_path, 'rb').read())
     io_obj.seek = None
     no_seek_io_embedding_objs = DeepFace.represent(io_obj)
-    assert defualt_embedding_objs == no_seek_io_embedding_objs
+    assert default_embedding_objs == no_seek_io_embedding_objs
 
     logger.info("✅ test standard represent with io object function done")
 
--- a/tests/test_represent.py
+++ b/tests/test_represent.py
@@ -34,7 +34,7 @@ def test_standard_represent_with_io_object():
 
     # Confirm non-image io objects raise exceptions
     with pytest.raises(ValueError, match='Failed to decode image'):
-        DeepFace.represent(io.BytesIO(open(__file__, 'rb').read()))
+        DeepFace.represent(io.BytesIO(open(r'../requirements.txt', 'rb').read()))
 
     logger.info("✅ test standard represent with io object function done")
 
--- a/deepface/DeepFace.py
+++ b/deepface/DeepFace.py
@@ -85,11 +85,13 @@ def verify(
     Verify if an image pair represents the same person or different persons.
     Args:
         img1_path (str or np.ndarray or IO[bytes] or List[float]): Path to the first image.
-            Accepts exact image path as a string, numpy array (BGR), base64 encoded images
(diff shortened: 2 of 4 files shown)


tests/test_represent.py
```diff
@@ -18,6 +19,21 @@ def test_standard_represent():
     logger.info("✅ test standard represent function done")
 
 
+def test_standard_represent_with_io_object():
+    img_path = "dataset/img1.jpg"
+    defualt_embedding_objs = DeepFace.represent(img_path)
```
serengil, line 24: typo

tests/test_represent.py
```diff
@@ -18,6 +19,21 @@ def test_standard_represent():
...
+    io_embedding_objs = DeepFace.represent(open(img_path, 'rb'))
```
serengil, line 25: what if i pass a text file as

```
io_obj = io.BytesIO(open("requirements.txt", 'rb').read())
```

tests/test_represent.py
PyWoody: You're too quick! I was just about to push the correction. It should be fixed now.

serengil:
TBH, i am not sure about this functionality.

Would you please create an issue first and let us to discuss first?

tests/test_represent.py
PyWoody: Yep, that'll work as `io.BytesIO` supports `.read`.

serengil: this is not okay then, we should be able to pass only image files to deepface functionalities.

PyWoody: Oh, I'm sorry, I misunderstood what you were asking.

An error would be raised by `np.frombuffer(obj.read(), np.uint8)` or `cv2.imdecode(nparr, cv2.IMREAD_COLOR)`, just as if you were trying to pass any other non-supported filetype as a string/filepath.


I only meant it would pass the `hasattr(img, "read") and callable(img.read)`

PyWoody:
Sure, I have to step out for a minute but I can create the issue when I get back.

tests/test_represent.py
serengil: I tried to call `load_image_from_io_object` with requirements.txt but it didn't throw any exception

serengil:
> Sure, I have to step out for a minute but I can create the issue when I get back.

No, not for this, before creating a PR. Otherwise, not merging comprehensive PR makes me sad.

PyWoody:
No worries. I'll make sure to do that next time.

It's no problem if you don't want to merge the PR. This is a feature I need for my personal use case anyways. I thought I should share in case others needed it, so there's no wasted effort.

tests/test_represent.py
PyWoody: I see. I had been following how `load_image_from_web` handled malformed images, but, if you want to raise the error in the function, I could update `load_from_io_object` to raise a ValueError if `cv2.imdecode` returns `None` like how `load_image_from_file_storage` is currently implemented.

It does appear modules in `deepface.modules` all already do `None` checks when loading the `img_path`, which `load_image_from_io_object` does follow in convention.

serengil: Would you please raise an error if it is not an image?

PyWoody: `load_image_from_io_object` now raises a `ValueException` if `cv2.imdecode` returns `None` for objects that aren't images, which is in line with how `load_image_from_file_storage` handles non-image objects.

serengil:
Would you please add this data type into the interface, too?

Argument data types and their docstrings

- https://github.com/serengil/deepface/blob/master/deepface/modules/detection.py#L22
- https://github.com/serengil/deepface/blob/master/deepface/DeepFace.py#L71
- https://github.com/serengil/deepface/blob/master/deepface/DeepFace.py#L167
- https://github.com/serengil/deepface/blob/master/deepface/DeepFace.py#L266
- https://github.com/serengil/deepface/blob/master/deepface/DeepFace.py#L372
- https://github.com/serengil/deepface/blob/master/deepface/DeepFace.py#L508

tests/test_represent.py
```diff
@@ -18,6 +20,25 @@ def test_standard_represent():
...
+
+    # Confirm non-seekable io objects are handled properly
+    io_obj = io.BytesIO(open(img_path, 'rb').read())
+    io_obj.seek = None
+    no_seek_io_embedding_objs = DeepFace.represent(io_obj)
+    assert default_embedding_objs == no_seek_io_embedding_objs
+
+    # Confirm non-image io objects raise exceptions
+    with pytest.raises(ValueError, match='Failed to decode image'):
+        DeepFace.represent(io.BytesIO(open(__file__, 'rb').read()))
```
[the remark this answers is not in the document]
PyWoody: Sure, should be all set now.

serengil:
> > Would you please add this data type into the interface, too?
> > Argument data types and their docstrings
> > 
> > * https://github.com/serengil/deepface/blob/master/deepface/modules/detection.py#L22
> > * https://github.com/serengil/deepface/blob/master/deepface/DeepFace.py#L71
> > * https://github.com/serengil/deepface/blob/master/deepface/DeepFace.py#L167
> > * https://github.com/serengil/deepface/blob/master/deepface/DeepFace.py#L266
> > * https://github.com/serengil/deepface/blob/master/deepface/DeepFace.py#L372
> > * https://github.com/serengil/deepface/blob/master/deepface/DeepFace.py#L508
> 
> Sorry, I should have done this at the outset. Before I go begin, do you agree that `IO[bytes]` is the correct type?

I think so, do you have any hesitation?

PyWoody:
Nope, that's how PIL does it and the others I checked. I just wanted to double-check first. These should be all set now.

serengil:
Would you please the types in docstring, too?

e.g. https://github.com/serengil/deepface/blob/master/deepface/DeepFace.py#L87

PyWoody:
`DeepFace.py` and `modules/detection.py` have had their respective docstrings updated.

PyWoody:
Ok, I'll update the descriptions. The current commit should fix the docstring issue. I didn't notice it regressed on my local during the `make lint` stage.

Issue #1426, opened 2025-01-14
[BUG]: Image format (RGB/BGR) is inconsistent when using DeepFace.represent


dakotah-jones:
### Before You Report a Bug, Please Confirm You Have Done The Following...

- [X] I have updated to the latest version of the packages.
- [X] I have searched for both [existing issues](https://github.com/serengil/deepface/issues) and [closed issues](https://github.com/serengil/deepface/issues?q=is%3Aissue+is%3Aclosed) and found none that matched my issue.

### DeepFace's version

v0.0.93

### Python version

3.12.8

### Operating System

Ubuntu 20.04.6 LTS

### Dependencies

N/A

### Reproducible example

```Python
representations = DeepFace.represent(
    file_path,
    model_name='Facenet',
    detector_backend='skip',
    normalization='Facenet'
)

# vs

representations = DeepFace.represent(
    file_path,
    model_name='Facenet',
    detector_backend='retinaface',
    normalization='Facenet'
)
```

### Relevant Log Output

N/A

### Expected Result

I expect the image format to be consistent regardless of `detector_backend`.

### What happened instead?

When `detector_backend` is `"skip"` the final image used for the embedding is in `"RGB"` format.
When `detector_backend` is any other valid option, like `"retinaface"` the final image used for the embedding is in `"BGR"` format.

### Additional Info

When using `detector_backend="skip"`:
- The image is loaded using `image_utils.load_image`. ([here](https://github.com/serengil/deepface/blob/29200f4fd6c330756481483e8702c1760d8c93cc/deepface/modules/representation.py#L89))
- The image is in `"BGR"` format.
- The image format is swapped using `img = img[:, :, ::-1]`. ([here](https://github.com/serengil/deepface/blob/29200f4fd6c330756481483e8702c1760d8c93cc/deepface/modules/representation.py#L120))
- The image is in `"RGB"` format.
- The image embedding is calculated.

When using any other `detector_backend`:
- `detection.extract_faces` is called.
- The `color_face` argument is defaulted to `"rgb"`
- The image is loaded using `image_utils.load_image`. ([here](https://github.com/serengil/deepface/blob/29200f4fd6c330756481483e8702c1760d8c93cc/deepface/modules/detection.py#L86))
- The image is now in `"BGR"` format.
[...]

PR #1427, opened 2025-01-15
Update representation.py

dakotah-jones:
Add explicit "color_face" argument.

## Tickets

https://github.com/serengil/deepface/issues/1426

### What has been done

With this PR I've added `color_face="bgr"` to `detection.extract_faces` call.

Previously: When using a `detector_backend` the final image sent to the model for embedding was `BGR"` format.
Now: When using a `detector_backend` the final image sent to the model for embedding is `"RGB"` format.

## How to test

```shell
make lint && make test
```


serengil:
I think we should change the skip case's behaviour, not default branch. All thresholds are tuned for default behaviour.

dakotah-jones:
I think that would depend on the expected `input` of `embedding = model.forward(img)`. ([here](https://github.com/serengil/deepface/blob/29200f4fd6c330756481483e8702c1760d8c93cc/deepface/modules/representation.py#L135))

Is the expected input to the model supposed to be **RGB** or **BGR** format?
Please clarify this for me.

If `img` is expected to be in **BGR** format, then I agree with you and I'll update the PR.
If `img` is expected to be in **RGB** format, then I disagree. The default case is incorrectly sending in `img` as **BGR**.

dakotah-jones:
The comments seem to suggest the input should be in **RGB** format. ([here](https://github.com/serengil/deepface/blob/29200f4fd6c330756481483e8702c1760d8c93cc/deepface/modules/representation.py#L119))

serengil:
Then, that comment should be fixed, too.

That line was updated in this PR. I missed it.

https://github.com/serengil/deepface/pull/1402

dakotah-jones:
For clarification sake.
Is the expected input to the model supposed to be **RGB** or **BGR** format?

It seems as though you are inferring that `img` sent for embedding should be in **BGR** format.

I ask because this is the opposite of my understanding.
I'm using this to generate FaceNet image embeddings. It's my understanding that FaceNet was trained on 160x160 images in **RGB** format.
It then seems strange that we would be calculating the embeddings on images in **BGR** format

dakotah-jones:
To be clear, I'm talking about the _**final**_ state of `img`, not the _**initial**_ state of the image loaded in.

serengil:
BGR

dakotah-jones:
Thank you for clarification.

The changes have been reverted.
The skip case has been updated.
The comment you mentioned from #1402 has been reverted.

serengil:
Did you run the tests after your change? Seems tests are broken.

dakotah-jones:
> BGR

Looking into the available `FacialRecognition` models.
It looks like some of the models used under the hood expect input as RGB and some expect input as BGR.

At the moment `represent` is always sending BGR format images to `model.forward`.

Let me know if this looks correct.

Expects BGR:
- VGG-Face
- OpenFace
- SFace
- GhostFaceNet

Expects RGB:
- FaceNet
- FaceNet512
- DeepFace
- DeepId
- Dlib
- ArcFace

serengil:
We are sending BGR to all models via model.forward

We are not doing anything with RGB.

dakotah-jones:
I understand the image being sent to model.forward is always BGR format.
That is a good thing. Having consistent input for `FacialRecognition` class implementation makes it easy to understand and use.

I'm not suggesting changing the input for `model.forward`.

I'm talking about the case where the underlying model has been trained on RGB images and we are sending in BGR images.

For instance, the `DlibClient` already has pre-processing steps to convert the image from BGR to RGB.
This is a case where you are using RGB.

I was just wondering if there were other cases, like `DlibClient`, where that type of pre-processing would be beneficial.

You may have a specific reason why most of these models take in BGR images as input. I would like to understand the reasoning behind that decision, for my own personal curiosity.

dakotah-jones:
On a more relevant note, those tests should pass now.
If you need anything else from me before approving that workflow, please let me know.

serengil:
This PR has some conflicts.

PR #1433, opened 2025-02-11
Batching on `.represent` to improve performance and utilize GPU in full

galthran-wq:
## Tickets

### What has been done

With this PR, `.represent` is able to accept a list of paths/numpy arrays and process them all in a batch.

## How to test

```shell
make lint && make test
```

I've made a collab notebook
https://colab.research.google.com/drive/1bV0yyrdT1a0a4dyemoeL5xf28w3ql1fd#scrollTo=lVndmFF5Kls7
which shows the >10x performance improvement and also the fact, that with batch_size=1 the GPU is almost not utilized

download

### DIFF
--- a/tests/test_represent.py
+++ b/tests/test_represent.py
@@ -88,14 +88,7 @@ def test_max_faces():
 @pytest.mark.parametrize("model_name", [
     "VGG-Face", 
     "Facenet", 
-    "Facenet512", 
-    "OpenFace", 
-    "DeepFace", 
-    "DeepID", 
-    "Dlib", 
-    "ArcFace", 
     "SFace", 
-    "GhostFaceNet"
 ])
 def test_batched_represent(model_name):
     img_paths = [
--- a/tests/test_represent.py
+++ b/tests/test_represent.py
@@ -3,6 +3,7 @@ import io
 import cv2
 import pytest
 import numpy as np
+import pytest
 
 # project dependencies
 from deepface import DeepFace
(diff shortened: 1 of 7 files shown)


tests/test_represent.py
```diff
@@ -81,3 +83,49 @@ def test_max_faces():
     max_faces = 1
     results = DeepFace.represent(img_path="dataset/couple.jpg", max_faces=max_faces)
     assert len(results) == max_faces
+
+
+@pytest.mark.parametrize("model_name", [
```
serengil, line 88: please do the with only one model - e.g. Facenet

serengil: okay but some are optional (e.g. dlib), these are not installed in github actions. so, your tests will be failed.

serengil:
There are some linting issues broken the actions

serengil:
LGTM

tests/test_represent.py
```diff
@@ -81,3 +83,42 @@ def test_max_faces():
...
+    "Facenet", 
+    "SFace", 
+])
+def test_batched_represent(model_name):
+    img_paths = [
+        "dataset/img1.jpg",
+        "dataset/img2.jpg",
+        "dataset/img3.jpg",
+        "dataset/img4.jpg",
+        "dataset/img5.jpg",
```
serengil, line 99: if you add couple.jpg here, there are 6 input images but in the response we will have 7 items.

the bad part, we cannot understand that which image has 2 faces. i am creating a PR to store input image's index in the response payload.

we may consider to have List of List of Dict response type for batch inputs in the future.

galthran-wq: I think it is a good idea to have List of List of Dict. I could make a PR now, or perhaps later, when batched detection is merged, because I also had in mind to use batched detection is `.represent` (now it is done in the for loop)

serengil: I will do the initial changes. PRs are always welcome!

serengil: i merged a workaround PR for this

galthran-wq: great work!

serengil: will optimize this because if batch size is long, this approach gives O(n^2) complexity.

I plan to do something with dict, which decrease the complexity O(n)

Issue #1434, opened 2025-02-12
[FEATURE]: Batching `.extract_faces`


galthran-wq:
### Description

Related: 
https://github.com/serengil/deepface/pull/1433
https://github.com/serengil/deepface/issues/1101

I think it is would be a good idea to be able to batch input images to detector.
The hypothesis is that it would improve the performance significantly.
Partially validate by https://github.com/serengil/retinaface/pull/116.

@serengil what do you think?

NatLee:
Hello, I have a question regarding `.extract_faces`. If it supports batch input, would the subsequent models, such as emotion or age detection, also need to be adjusted based on the output of `.extract_faces`?

For instance, if I have three input images, each will produce face detection results. How should the subsequent analysis define multiple faces across these three images? Perhaps using the input order to each face could be a solution?

galthran-wq:
hi. I do not think they would *have to be* adjusted, because the function is supposed to work exactly the same for a single image input and all the susequent models are using it currently with a single image. 

Now, if we were to adjust the usage in subsequent models to batched inputs, then you indeed raise a good point on how to recover which detection results correspond to which images in the input. And I think the solution here is this. The output type of the `.extract_faces` should be `Union[List[Dict[str, Any]], List[List[Dict[str, Any]]`. So, it is `List[Dict[str, Any]]` for a single image input (just like before), and `List[List[Dict[str, Any]` if a list of images is passed -- for each input face, a list of detected faces.

I'm planing to incorporate this change in the PR

serengil:
@galthran-wq so, if you feed 3 images to input as list or numpy and the 2nd one has 5 faces, then function will still return a list of 3. The second item in the response will have 5 items. Liked that design.

sangeethkc:
**Memory Issue with `extract_faces` on Large Images**  

I'm experiencing a memory issue when using `extract_faces` on an image containing 66 faces. The process consumed about **86% of memory** on a **32GB server** and took **over 4 minutes** to complete.  

I used the original image because resizing caused accuracy issues. The **RetinaFace** backend provided the most accurate count.  

### **Test Code**  
```python
detector_backends = ['opencv', 'retinaface'] 
aligns = [True, False]

for detector_backend in detector_backends:
    for align in aligns:
        print(f"---------------{detector_backend} start-----------------")
        
        source_image_faces = preprocess_image(img='./images/s.jpg', detector_backend=detector_backend)
[...]

PR #1435, opened 2025-02-13
Batching on `.extract_faces` to improve performance and utilize GPU in full

galthran-wq:
## Tickets

https://github.com/serengil/deepface/pull/1433
https://github.com/serengil/deepface/issues/1434

### What has been done

With this PR, `.extract_faces` is able to accept a list of images

## How to test

```shell
make lint && make test
```

Benchmarking on detecting 50 faces:
image

For yolov11n, batch size 20 is 59.27% faster than batch size 1.
For yolov11s, batch size 20 is 29.00% faster than batch size 1.
For yolov11m, batch size 20 is 31.73% faster than batch size 1.
For yolov8, batch size 20 is 12.68% faster than batch size 1.

### DIFF
--- a/deepface/models/Detector.py
+++ b/deepface/models/Detector.py
@@ -1,7 +1,36 @@
 from typing import List, Tuple, Optional, Union
-from abc import ABC
+from abc import ABC, abstractmethod
 from dataclasses import dataclass
 import numpy as np
+
+# Notice that all facial detector models must be inherited from this class
+
+
+# pylint: disable=unnecessary-pass, too-few-public-methods, too-many-instance-attributes
+class Detector(ABC):
+    @abstractmethod
+    def detect_faces(
+        self,
+        img: Union[np.ndarray, List[np.ndarray]]
+    ) -> Union[List["FacialAreaRegion"], List[List["FacialAreaRegion"]]]:
+        """
+        Interface for detect and align faces in a batch of images
+
+        Args:
+            img (Union[np.ndarray, List[np.ndarray]]): 
+            Pre-loaded image as numpy array or a list of those
+
+        Returns:
+            results (Union[List[List[FacialAreaRegion]], List[FacialAreaRegion]]): 
+            A list or a list of lists of FacialAreaRegion objects
+                where each object contains:
+
+            - facial_area (FacialAreaRegion): The facial area region represented
+                as x, y, w, h, left_eye and right_eye. left eye and right eye are
+                eyes on the left and right respectively with respect to the person
+                instead of observer.
+        """
+        pass
 
 
 # pylint: disable=unnecessary-pass, too-few-public-methods, too-many-instance-attributes
@@ -49,56 +78,3 @@ class DetectedFace:
     img: np.ndarray
     facial_area: FacialAreaRegion
     confidence: float
-
-
-# Notice that all facial detector models must be inherited from this class
-
-class Detector(ABC):
-
-    def detect_faces(
-        self,
-        img: Union[np.ndarray, List[np.ndarray]],
-    ) -> Union[List[FacialAreaRegion], List[List[FacialAreaRegion]]]:
-        """
-        Detect and align faces in an image or a list of images
-
-        Args:
-            img (Union[np.ndarray, List[np.ndarray]]): 
-            pre-loaded image as numpy array or a list of those
-
-        Returns:
-            results (Union[List[FacialAreaRegion], List[List[FacialAreaRegion]]]): 
-            A list or a list of lists of FacialAreaRegion objects
-        """
-        is_batched_input = isinstance(img, list)
-        if not is_batched_input:
-            img = [img]
-        results = [self._process_single_image(single_img) for single_img in img]
-        if not is_batched_input:
-            return results[0]
-        return results
-
-    def _process_single_image(
-        self,
-        img: np.ndarray
-    ) -> List[FacialAreaRegion]:
-        """
-        Interface for detect and align faces in a single image
-
-        Args:
-            img (Union[np.ndarray, List[np.ndarray]]): 
-            Pre-loaded image as numpy array or a list of those
-
-        Returns:
-            results (Union[List[List[FacialAreaRegion]], List[FacialAreaRegion]]): 
-            A list or a list of lists of FacialAreaRegion objects
-                where each object contains:
-
-            - facial_area (FacialAreaRegion): The facial area region represented
-                as x, y, w, h, left_eye and right_eye. left eye and right eye are
-                eyes on the left and right respectively with respect to the person
-                instead of observer.
-        """
-        raise NotImplementedError(
-            "Subclasses that do not implement batch detection must implement this method"
-        )
--- a/deepface/models/Detector.py
+++ b/deepface/models/Detector.py
@@ -10,14 +10,15 @@ import numpy as np
 class Detector(ABC):
     @abstractmethod
     def detect_faces(
-        self, 
-        imgs: Union[np.ndarray, List[np.ndarray]]
+        self,
+        img: Union[np.ndarray, List[np.ndarray]]
     ) -> Union[List["FacialAreaRegion"], List[List["FacialAreaRegion"]]]:
         """
         Interface for detect and align faces in a batch of images
 
         Args:
-            imgs (Union[np.ndarray, List[np.ndarray]]): pre-loaded image as numpy array or a list of those
+            img (Union[np.ndarray, List[np.ndarray]]): 
+            Pre-loaded image as numpy array or a list of those
 
         Returns:
             results (Union[List[List[FacialAreaRegion]], List[FacialAreaRegion]]): 
--- a/deepface/models/Detector.py
+++ b/deepface/models/Detector.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass
 import numpy as np
 
 
+# pylint: disable=unnecessary-pass, too-few-public-methods, too-many-instance-attributes
 @dataclass
 class FacialAreaRegion:
     """
@@ -52,7 +53,6 @@ class DetectedFace:
 
 # Notice that all facial detector models must be inherited from this class
 
-# pylint: disable=unnecessary-pass, too-few-public-methods, too-many-instance-attributes
 class Detector(ABC):
 
     def detect_faces(
--- a/deepface/models/Detector.py
+++ b/deepface/models/Detector.py
@@ -1,36 +1,7 @@
 from typing import List, Tuple, Optional, Union
-from abc import ABC, abstractmethod
+from abc import ABC
 from dataclasses import dataclass
 import numpy as np
-
-# Notice that all facial detector models must be inherited from this class
-
-
-# pylint: disable=unnecessary-pass, too-few-public-methods, too-many-instance-attributes
-class Detector(ABC):
-    @abstractmethod
-    def detect_faces(
-        self,
-        img: Union[np.ndarray, List[np.ndarray]]
-    ) -> Union[List["FacialAreaRegion"], List[List["FacialAreaRegion"]]]:
-        """
-        Interface for detect and align faces in a batch of images
-
-        Args:
-            img (Union[np.ndarray, List[np.ndarray]]): 
-            Pre-loaded image as numpy array or a list of those
-
-        Returns:
-            results (Union[List[List[FacialAreaRegion]], List[FacialAreaRegion]]): 
-            A list or a list of lists of FacialAreaRegion objects
-                where each object contains:
-
-            - facial_area (FacialAreaRegion): The facial area region represented
-                as x, y, w, h, left_eye and right_eye. left eye and right eye are
-                eyes on the left and right respectively with respect to the person
-                instead of observer.
-        """
-        pass
 
 
 @dataclass
@@ -77,3 +48,57 @@ class DetectedFace:
     img: np.ndarray
     facial_area: FacialAreaRegion
     confidence: float
+
+
+# Notice that all facial detector models must be inherited from this class
+
+# pylint: disable=unnecessary-pass, too-few-public-methods, too-many-instance-attributes
+class Detector(ABC):
+
+    def detect_faces(
+        self,
+        img: Union[np.ndarray, List[np.ndarray]],
+    ) -> Union[List[FacialAreaRegion], List[List[FacialAreaRegion]]]:
+        """
+        Detect and align faces in an image or a list of images
+
+        Args:
+            img (Union[np.ndarray, List[np.ndarray]]): 
+            pre-loaded image as numpy array or a list of those
+
+        Returns:
+            results (Union[List[FacialAreaRegion], List[List[FacialAreaRegion]]]): 
+            A list or a list of lists of FacialAreaRegion objects
+        """
+        is_batched_input = isinstance(img, list)
+        if not is_batched_input:
+            img = [img]
+        results = [self._process_single_image(single_img) for single_img in img]
+        if not is_batched_input:
+            return results[0]
+        return results
+
+    def _process_single_image(
+        self,
+        img: np.ndarray
+    ) -> List[FacialAreaRegion]:
+        """
+        Interface for detect and align faces in a single image
+
+        Args:
+            img (Union[np.ndarray, List[np.ndarray]]): 
+            Pre-loaded image as numpy array or a list of those
+
+        Returns:
+            results (Union[List[List[FacialAreaRegion]], List[FacialAreaRegion]]): 
+            A list or a list of lists of FacialAreaRegion objects
+                where each object contains:
+
+            - facial_area (FacialAreaRegion): The facial area region represented
+                as x, y, w, h, left_eye and right_eye. left eye and right eye are
+                eyes on the left and right respectively with respect to the person
+                instead of observer.
+        """
+        raise NotImplementedError(
+            "Subclasses that do not implement batch detection must implement this method"
+        )
--- a/deepface/models/Detector.py
+++ b/deepface/models/Detector.py
@@ -1,4 +1,4 @@
-from typing import List, Tuple, Optional
+from typing import List, Tuple, Optional, Union
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 import numpy as np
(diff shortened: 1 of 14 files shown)


skyler14:
Do you have a branch in your fork that currently combines all the optimizations you've submitted. I'd like to start using them while the approval process is going

whats been the total speedup you've been able to see

galthran-wq:
I do. You can check
https://github.com/galthran-wq/deepface/tree/master-enhanced 

it combines these two PRs with some other small modifications:
- `.represent` uses batched detector inference (here https://github.com/serengil/deepface/pull/1433 it only does batched embedding, because batched detection is not yet implemented)
- `.represent` returns a list of list of dicts, if a batch of images is passed. This is neccessary to be able to recover to which images the resulting faces correspond to. It might be a good idea to include this change in the PR as well. You can check the test in the fork.

Not all of the detectors currently (both in this PR and in the fork) implement batching. In particular, YOLO does. I've found it to be optimal in terms of performance and inference speed. The only problem is installing both torch and tensorflow with GPU, but I've managed to somehow do that.


All in all, with the combination of `yolov11m` and `Facenet`, both using GPU, and batch size 100 (the largest I could fit in 4090) I am seeing aroung 15x speed boost, but that is highly dependent on the input images, the GPU (especially memory size). I've also had a quick peek and it seems like the performance on the CPU is improved as well.


@serengil FYI I would be happy to contribute the aforementioned modifications if we have progress on the PRs.

serengil:
I will review this PR this week i hope

serengil:
Seems this breaks the unit tests. Must be sorted.

galthran-wq:
should be good now

serengil:
Nope, still failing.

tests/test_extract_faces.py
```diff
@@ -79,6 +79,50 @@ def test_different_detectors():
         logger.info(f"✅ extract_faces for {detector} backend test is done")
 
 
+@pytest.mark.parametrize("detector_backend", [
+    "opencv",
+    "ssd"
+])
+def test_batch_extract_faces(detector_backend):
+    img_paths = [
```
serengil, line 87: do you mind to add couple.jpg here? it has 2 faces. so, result should be list of 4 items and for the one coming from couples.jpg should have many items.

tests/test_extract_faces.py
```diff
@@ -79,6 +79,50 @@ def test_different_detectors():
         logger.info(f"✅ extract_faces for {detector} backend test is done")
 
 
+@pytest.mark.parametrize("detector_backend", [
+    "opencv",
```
serengil, line 83: please add mtcnn, retinaface and all non-optional ones

deepface/models/face_detection/RetinaFace.py
```diff
@@ -13,64 +13,72 @@ class RetinaFaceClient(Detector):
...
-                confidence=confidence,
-                nose=nose,
-                mouth_left=mouth_left,
-                mouth_right=mouth_right,
-            )
-
-            resp.append(facial_area)
-
-        return resp
+        if isinstance(img, np.ndarray):
```
serengil, line 31: what if img is (2, 224, 224, 3) sized numpy array? i mean many images in single numpy array.

serengil:
You implemented OpenCv, Ssd, Yolo, MtCnn and RetinaFace to accept list inputs

What if I send list to YuNet, MediaPipe, FastMtCnn, Dlib or CenterFace?

I assume an exception will be thrown, but users should see a meaningful message.

tests/test_extract_faces.py
```diff
@@ -79,6 +79,144 @@ def test_different_detectors():
...
+    )
+
+    # compare results
+    for img_objs_batch, img_objs_batch_paths in zip(imgs_objs_batch, imgs_objs_batch_paths):
+        assert len(img_objs_batch) == len(img_objs_batch_paths), (
+            "Batch and individual extraction results should have the same number of detected faces"
+        )
+
+
+def test_batch_extract_faces_single_image():
```
galthran-wq, line 210: FYI this might be not the expected behaviour

serengil:
@galthran-wq you are pushing too many things, would you please inform me when it is ready.

tests/test_extract_faces.py
```diff
@@ -79,6 +79,144 @@ def test_different_detectors():
...
+    "centerface",
+    # optional
+    # "yolov11s",
+    # "mediapipe",
+    # "dlib",
+])
+def test_batch_extract_faces(detector_backend):
+    detector_backend_to_rtol = {
+        "opencv": 0.1,
+        "mtcnn": 0.2,
```
galthran-wq, line 97: turns out batching has different impact on the result for different backends. It is mostly the same (for almost all the models, the relative error is <1%.
mtcnn and opencv suffer a bit more, not quite sure why (which features are different exactly)

tests/test_extract_faces.py
```diff
@@ -79,6 +79,144 @@ def test_different_detectors():
...
 
 
+@pytest.mark.parametrize("detector_backend", [
+    "opencv",
+    "ssd",
+    "mtcnn",
+    "retinaface",
+    "yunet",
+    "centerface",
+    # optional
```
galthran-wq, line 89: this might be a good idea to include requirements_additional.txt in the test action so that we could test those as well

serengil: nope, because unit tests will take 1 hour then

galthran-wq:
> @galthran-wq you are pushing too many things, would you please inform me when it is ready.

it is ready now, im usually pushing when it's done

I've just fixed what you've noted:
- added psedu-batching (for loop) for other models
- support for np array
- changed detect_faces to return a list of list on batched input
- couple more tests

tests/test_extract_faces.py
serengil: we may consider to throw exception for those detectors as they are not suitable for batch maybe

tests/test_extract_faces.py
```diff
@@ -79,6 +79,144 @@ def test_different_detectors():
...
+        img_path=img_paths,
+        detector_backend=detector_backend,
+        align=True,
+    )
+    
+    assert (
+        len(imgs_objs_batch) == 4 and 
+        all(isinstance(obj, list) for obj in imgs_objs_batch)
+    )
+    assert all(
```
serengil, line 128: please add comment, the last one has many faces, others have just one face

tests/test_extract_faces.py
serengil: what do you mean?

tests/test_extract_faces.py
```diff
@@ -79,6 +79,144 @@ def test_different_detectors():
         logger.info(f"✅ extract_faces for {detector} backend test is done")
 
 
+@pytest.mark.parametrize("detector_backend", [
```
serengil, line 82: please add an unit test case for

```python
img1 = cv2.imread(img1_path)
img2 = cv2.imread(img2_path)

img = np.stack([img1, img2])

assert len(img.shape) == 4 # Check dimension.
assert img.shape[0] == 2 # Check batch size.
```

what if image is batch in numpy format?

deepface/models/face_detection/OpenCv.py
```diff
@@ -29,55 +29,63 @@ def build_model(self):
...
 
         Returns:
-            results (List[FacialAreaRegion]): A list of FacialAreaRegion objects
+            results (Union[List[FacialAreaRegion], List[List[FacialAreaRegion]]]): 
+            A list or a list of lists of FacialAreaRegion objects
         """
-        resp = []
-
-        detected_face = None
+        if isinstance(img, np.ndarray):
```
serengil, line 47: I recommend you to call detect_faces itself it if it is list. this can also be done in Detector.py or detection.py.

in that way, we will not change all detectors

serengil:
is this ready? i can still see my comments not resolved.

tests/test_extract_faces.py
```diff
@@ -79,6 +79,144 @@ def test_different_detectors():
...
+    imgs_objs_individual = [
+        DeepFace.extract_faces(
+            img_path=img_path,
+            detector_backend=detector_backend,
+            align=True,
+        ) for img_path in img_paths
+    ]
+    
+    # Extract faces in batch
+    imgs_objs_batch = DeepFace.extract_faces(
```
serengil, line 118: please check the return type

when it is batch input, then return type should be List[List[Dict[str, Any]]

similarly, put return type check for non batch input test, too - in test_different_detectors case

tests/test_extract_faces.py
```diff
@@ -79,6 +79,144 @@ def test_different_detectors():
...
+    
+    assert (
+        len(imgs_objs_batch) == 4 and 
+        all(isinstance(obj, list) for obj in imgs_objs_batch)
+    )
+    assert all(
+        len(imgs_objs_batch[i]) == 1 
+        for i in range(len(imgs_objs_batch[:-1]))
+    )
+    assert len(imgs_objs_batch[-1]) == 2
```
serengil, line 132: it is hard to understand why the last one should have 2 and others should have 1

recommend to store expected_num_faces list - https://github.com/serengil/deepface/blob/master/tests/test_represent.py#L212

tests/test_extract_faces.py
```diff
@@ -79,6 +79,144 @@ def test_different_detectors():
...
+    imgs = [
+        cv2.resize(image_utils.load_image(img_path)[0], (1920, 1080))
+        for img_path in img_paths
+    ]
+
+    # load images as numpy arrays
+    imgs_batch = np.stack(imgs, axis=0)
+
+    # extract faces in batch of numpy arrays
+    imgs_objs_batch = DeepFace.extract_faces(
```
serengil, line 188: check the size of imgs_objs_batch - it should be 4 - len(img_paths)

also last item should have 2 items

galthran-wq:
To summarize what's changed:

- I've added comments and additional checks to the tests.
- I've made batching on `opencv` and `mtcnn` optional (due to the above issue). To enforce batching a user can set `ENABLE_OPENCV_BATCH_DETECTION` (or `ENABLE_MTCNN_BATCH_DETECTION`) to `true`.

Unfortunately, this didn't fix the batch extraction case on `opencv` -- the problem is that is *occasionally* fails (it seems that the predictions might be different). Note that it has nothing to do with batching, because it is disabled by default. We might add a separate issue and test to reproduce this.

- i have fixed the special case of a single image in a list input (batch of size 1). It now indeed returns a list with a single element -- the list of detected faces in that image.
- those detectors that do no implement batching all had repeating logic in `detect_faces`. I have moved this logic to the default implementation in `Detector`. Now, those detectors only need to implement `_process_single_image`, and batching would be supported by inheritance.
- If a detector implements batching, then it overrides `detect_faces` with this logic, just as before.

tests/test_extract_faces.py
```diff
@@ -79,6 +83,169 @@ def test_different_detectors():
...
+        isinstance(imgs_objs_batch, list) and 
+        all(
+            isinstance(obj, list) and 
+            all(isinstance(face, dict) for face in obj) 
+            for obj in imgs_objs_batch
+        )
+    )
+
+    # Check that the batch extraction returned the expected number of face lists
+    assert len(imgs_objs_batch) == 4
```
serengil, line 220: instead of it is equal to 4, check it is equal to len(img_paths)

otherwise if i add a new image to img_path, tests will fail

tests/test_extract_faces.py
serengil: what do you mean here?

galthran-wq: I just wanted to draw your attention to the fact that previously, if a single image is passed in a list, it was interpreted as not batched input, so that a list of faces is returned (not a list of lists).
It is not the case in the current state ofthe code, it was fixed.
So this test currently just sanity checks this special case

deepface/modules/detection.py
```diff
@@ -80,135 +82,158 @@ def extract_faces(
             just available in the result only if anti_spoofing is set to True in input arguments.
     """
 
-    resp_objs = []
+    batched_input = (
+        (
+            isinstance(img_path, np.ndarray) and
+            img_path.ndim == 4
```
serengil, line 88: `isinstance(img_path, np.ndarray) and img_path.ndim == 4 and img_path.shape[0] == 1` -> this is not a batched input, right?

deepface/models/Detector.py
```diff
@@ -72,3 +49,56 @@ class DetectedFace:
...
+            A list or a list of lists of FacialAreaRegion objects
+                where each object contains:
+
+            - facial_area (FacialAreaRegion): The facial area region represented
+                as x, y, w, h, left_eye and right_eye. left eye and right eye are
+                eyes on the left and right respectively with respect to the person
+                instead of observer.
+        """
+        raise NotImplementedError(
+            "Subclasses that do not implement batch detection must implement this method"
```
serengil, line 103: it is not clear that which subclass is giving this error. IMO, we should print the detector name in the exception message.

serengil:
I don't like this approach:

```
those detectors that do no implement batching all had repeating logic in detect_faces. I have moved this logic to the default implementation in Detector. Now, those detectors only need to implement _process_single_image, and batching would be supported by inheritance.
If a detector implements batching, then it overrides detect_faces with this logic, just as before.
```

all detectors should have `detect_faces` instead of `_process_single_image`, and you should raise an error in that detector if it is not supported.

galthran-wq:
Changes:

1. reverted `_process_single_image`
2. disabled option to use batch mode for mtcnn and opencv

so now they just use a for loop to process all the images, if a batch is passed. A user gets a warning that there is no actual batching happening.

Do you still think it is a good idea to raise an error?

deepface/models/face_detection/MtCnn.py
```diff
@@ -1,5 +1,6 @@
 # built-in dependencies
-from typing import List
+import logging
```
serengil, line 2: we are not using logging module directly anywhere. use this instead:

```python
from deepface.commons.logger import Logger
logger = Logger()
```

deepface/models/face_detection/MtCnn.py
```diff
@@ -8,6 +9,8 @@
 # project dependencies
 from deepface.models.Detector import Detector, FacialAreaRegion
 
+logger = logging.getLogger(__name__)
```
serengil, line 12: logger = Logger()

deepface/models/face_detection/OpenCv.py
```diff
@@ -1,6 +1,7 @@
 # built-in dependencies
 import os
-from typing import Any, List
+from typing import Any, List, Union
+import logging
```
serengil, line 4: same here.

deepface/models/face_detection/RetinaFace.py
```diff
@@ -13,64 +13,75 @@ class RetinaFaceClient(Detector):
...
-                confidence=confidence,
-                nose=nose,
-                mouth_left=mouth_left,
-                mouth_right=mouth_right,
-            )
-
-            resp.append(facial_area)
-
-        return resp
+        is_batched_input = isinstance(img, list)
```
serengil, line 31: you are writing same code in every detector

```python
        is_batched_input = isinstance(img, list)
        if not is_batched_input:
            imgs = [img]
        else:
            imgs = img
```

IMO, this should be done in the parent where those detectors are being called. In that way, we will not have repeated code.

deepface/models/face_detection/RetinaFace.py
```diff
@@ -13,64 +13,75 @@ class RetinaFaceClient(Detector):
...
+                        nose=nose,
+                        mouth_left=mouth_left,
+                        mouth_right=mouth_right,
+                    )
+
+                    resp.append(facial_area)
+
+            batch_results.append(resp)
+
+        if not is_batched_input:
```
serengil, line 85: same here, we can do this in parent and avoid repeated code.

```python
        if not is_batched_input:
            return batch_results[0]
        return batch_results
```

deepface/models/face_detection/MtCnn.py
```diff
@@ -16,45 +19,71 @@ class MtCnnClient(Detector):
...
-                    w=w,
-                    h=h,
-                    left_eye=left_eye,
-                    right_eye=right_eye,
-                    confidence=confidence,
-                )
-
-                resp.append(facial_area)
+        img_rgb = [img[:, :, ::-1] for img in img]
+        if self.supports_batch_detection:
```
serengil, line 50: self.supports_batch_detection is returning false always. so, get rid of _supports_batch_detection.

have just this

```python
detections = [self.model.detect_faces(single_img) for single_img in img_rgb]
```

deepface/models/face_detection/OpenCv.py
```diff
@@ -29,55 +32,72 @@ def build_model(self):
         detector["eye_detector"] = self.__build_cascade("haarcascade_eye")
         return detector
 
-    def detect_faces(self, img: np.ndarray) -> List[FacialAreaRegion]:
+    def _supports_batch_detection(self) -> bool:
```
serengil, line 35: no need _supports_batch_detection, it is returning false always

deepface/models/face_detection/OpenCv.py
```diff
@@ -29,55 +32,72 @@ def build_model(self):
...
-            results (List[FacialAreaRegion]): A list of FacialAreaRegion objects
+            results (Union[List[FacialAreaRegion], List[List[FacialAreaRegion]]]): 
+            A list or a list of lists of FacialAreaRegion objects
         """
-        resp = []
-
-        detected_face = None
+        if isinstance(img, np.ndarray):
+            imgs = [img]
+        elif self.supports_batch_detection:
```
serengil, line 59: this condition is always false, get rid of it

tests/test_extract_faces.py
```diff
@@ -79,6 +83,169 @@ def test_different_detectors():
         logger.info(f"✅ extract_faces for {detector} backend test is done")
 
 
+@pytest.mark.parametrize("detector_backend", [
+    # "opencv",
```
serengil, line 87: activate opencv please, it is default detector. important to have tests.

tests/test_extract_faces.py
```diff
@@ -79,6 +83,169 @@ def test_different_detectors():
...
+        "dataset/couple.jpg"
+    ]
+    imgs = [
+        cv2.resize(image_utils.load_image(img_path)[0], (1920, 1080))
+        for img_path in img_paths
+    ]
+    expected_num_faces = [1, 1, 1, 2]
+
+    # load images as numpy arrays
+    imgs_batch = np.stack(imgs, axis=0)
```
serengil, line 199: please add this control

`assert imgs_batch.ndim == 4 and imgs_batch.shape[0] == 4`

tests/test_extract_faces.py
```diff
@@ -79,6 +83,169 @@ def test_different_detectors():
...
+        )
+
+
+def test_batch_extract_faces_single_image():
+    img_path = "dataset/couple.jpg"
+    imgs_objs_batch = DeepFace.extract_faces(
+        img_path=[img_path],
+        align=True,
+    )
+    assert len(imgs_objs_batch) == 1 and isinstance(imgs_objs_batch[0], list)
```
serengil, line 245: please check this

assert len(imgs_objs_batch[0]) == 2

serengil:
> so now they just use a for loop to process all the images, if a batch is passed. A user gets a warning that there is no actual batching happening.

for loop is okay, don't raise error

serengil:
As I understand, you built a for loop in each detector. Is that really a batch?

What if we build a for loop in the parent of detectors, and don't change detectors?

serengil:
I mean batch is really running for yolo

We can create a method extract faces from batch in parent class of detectors, and only customize this for yolo.

In parent class, we basically build a for loop and call detector

In that way, detectors will not change

serengil:
this has been inactive for a long time, will close it if no objection

Issue #1187, opened 2024-04-14
Possible to add InsightFace


ethaniel:
Hi! Thank you for the amazing library, it really was a gentle introduction into the world of facial recognition.

However, I've noticed that none of the models really work for my face, I get false and false positive results (I've tried different detectors and normalizers too). I've also noticed that the models that you've linked are all around 4 years old. 

Perhaps, adding some newer models (like InsightFace) would be possible?

serengil:
ArcFace is the facial recognition model of insightface project.

serengil:
Besides, GhostFace is a recently published model.

ethaniel:
@serengil thank you! One small thing - looks like Arcface has released new models ([r100,AdaFace](https://github.com/leondgarse/Keras_insightface?tab=readme-ov-file#current-accuracy) back in 2022. Is that the one that you [ported](https://sefiks.com/2020/12/14/deep-face-recognition-with-arcface-in-keras-and-python/) to deepface?

ethaniel:
Thank you! Any chance that [r100,AdaFace](https://github.com/leondgarse/Keras_insightface/releases/download/v1.0.0/TT_r100_max_pool_E_prelu_dr04_lr_01_l2_5e4_adaface_emb512_sgd_m09_bs512_glint360k_64_only_margin_SG_scale_true_bias_false_random_100_partial4_basic_model_latest.h5) could be ported into your library?

I've got an asian face and no matter how much I try, I couldn't get Arcface (or any other model in the library) to provide correct embeddings that would match me on different photos through the last 10 years. I was hoping that one of the newer models from Arcface could have possibly fixed that.

serengil:
of course, will try to add AdaFace model as well but its lfw score is very close, do not think it will resolve your problem.

ethaniel:
InsightFace have also been experimenting with south asian and east asian faces (https://github.com/deepinsight/insightface/tree/master/model_zoo), getting good results on their R100,Glint360K model. However that's onnx format and waay over my head to understand if it's even possible to be ported into deepface.

Sefik, I just wanted to say that I appreciate you and everything that you do! Thank you so so much for opening the world of facial recognition for people that want to learn!

serengil:
Which onnx fits you?

ethaniel:
Looking again at that list, "buffalo_l" model is the most promising. It has the best South Asian (93.16) and East Asian (74.96) accuracy of all + the LFW is at 99.83.

[...]

PR #1437, opened 2025-02-18
added a test file for buffalo-l model of insightface

Raghucharan16:
## Tickets

https://github.com/serengil/deepface/issues/1187

### What has been done

With this PR, This is not a complete functionality yet, code runs without any errors and if we want to make it work in deepface, may be we need to bypass face detection and simply pass the images to the model. when i implemented Buffalo_l class follwed by Facialrecognition class, while forwarding, in the initializing phase, i'm getting no face detected error. 

## How to test

```shell
python tests/test_buffalo_l.py
```
for full implementation i may need assistance of you @serengil

### DIFF
--- a/tests/test_buffalo_l
+++ b/tests/test_buffalo_l
@@ -1,0 +1,46 @@
+import cv2
+import insightface
+import numpy as np
+from insightface.app import FaceAnalysis
+
+# Initialize face analysis model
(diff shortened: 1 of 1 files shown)


serengil:
This is not a real implementation, isn't it?

Raghucharan16:
@serengil indeed that is the real implementation. You can check the insightface library. I myself have added the example implementation to their example page.

Raghucharan16:
I don't want to merge this directly and need assistance to hone the implementation according to deepface..

PR #1446, opened 2025-02-26
allow 0 (as integers) when using pre-calculated embeddings in verify

jabiinfante:
## Tickets

(No ticket has been opened since it's a small edge case.)

First of all, I'd like to thank you for the amazing work you've done with DeepFace! It's an incredibly useful library so far with incredible successful results.

While working with pre-calculated embeddings and calling the DeepFace API from JavaScript, I encountered a small issue. Since JavaScript does not allow the creation of JSON objects with 0.0 (floating point 0), I needed to update the check in the code to allow 0 as a valid value alongside float values.

### What has been done

- Iterate once over img_path (as done before with _all_)
- Check if each value is either a float or 0.
- Convert the values to float as needed, including 0 (which will be converted to 0.0).

## How to test

```shell
make lint && make test
```

### DIFF
--- a/deepface/modules/verification.py
+++ b/deepface/modules/verification.py
@@ -141,7 +141,8 @@ def verify(
         """
         if isinstance(img_path, list):
             # given image is already pre-calculated embedding
-            if not all(isinstance(dim, float) for dim in img_path):
+            if not all(isinstance(dim, (float, int)) for dim in img_path):
+
                 raise ValueError(
                     f"When passing img{index}_path as a list,"
                     " ensure that all its items are of type float."


deepface/modules/verification.py
```diff
@@ -141,11 +141,16 @@ def extract_embeddings_and_facial_areas(
         """
         if isinstance(img_path, list):
             # given image is already pre-calculated embedding
-            if not all(isinstance(dim, float) for dim in img_path):
-                raise ValueError(
-                    f"When passing img{index}_path as a list,"
-                    " ensure that all its items are of type float."
-                )
+            validated_pre_calculated_embeddings = []
```
serengil, line 144: When you send 0, i think python thinks it as int instead of float and that was the issue i suppose.

to have a clearer solution, we may let dim be int or float

```python
if not all(isinstance(dim, (float, int)) for dim in img_path):
```

what do you think?

jabiinfante: That was my initial approach (and it worked as well).

However, I added this:
```python
 img_path = list(map(float, img_path))
```
This way, everything was converted to float again, and it seemed to me like an unnecessary iteration.

I just gave it a quick try as you suggested (without any further casting), and all tests passed, and my local tests seem fine.

serengil:
LGTM, thank you for your contribution

PR #1447, opened 2025-02-27
feat: add missing api options

charsleysa:
## Tickets

N/A

### What has been done

Added missing options to API endpoints, such as `expand_percentage`, `normalization`, and `threshold`.

## How to test

```shell
make lint && make test
```

### DIFF
--- a/deepface/api/src/modules/core/routes.py
+++ b/deepface/api/src/modules/core/routes.py
@@ -87,6 +87,8 @@ def represent():
         detector_backend=input_args.get("detector_backend", "opencv"),
         enforce_detection=input_args.get("enforce_detection", True),
         align=input_args.get("align", True),
+        expand_percentage=input_args.get("expand_percentage", 0),
+        normalization=input_args.get("normalization", "base"),
         anti_spoofing=input_args.get("anti_spoofing", False),
         max_faces=input_args.get("max_faces"),
     )
@@ -118,8 +120,12 @@ def verify():
         model_name=input_args.get("model_name", "VGG-Face"),
         detector_backend=input_args.get("detector_backend", "opencv"),
         distance_metric=input_args.get("distance_metric", "cosine"),
-        align=input_args.get("align", True),
         enforce_detection=input_args.get("enforce_detection", True),
+        align=input_args.get("align", True),
+        expand_percentage=input_args.get("expand_percentage", 0),
+        normalization=input_args.get("normalization", "base"),
+        silent=input_args.get("silent", False),
+        threshold=input_args.get("threshold", None),
         anti_spoofing=input_args.get("anti_spoofing", False),
     )
 
@@ -160,6 +166,8 @@ def analyze():
         detector_backend=input_args.get("detector_backend", "opencv"),
         enforce_detection=input_args.get("enforce_detection", True),
         align=input_args.get("align", True),
+        expand_percentage=input_args.get("expand_percentage", 0),
+        silent=input_args.get("silent", False),
         anti_spoofing=input_args.get("anti_spoofing", False),
     )
 
--- a/deepface/api/src/modules/core/routes.py
+++ b/deepface/api/src/modules/core/routes.py
@@ -85,12 +85,12 @@ def represent():
         img_path=img,
         model_name=input_args.get("model_name", "VGG-Face"),
         detector_backend=input_args.get("detector_backend", "opencv"),
-        enforce_detection=input_args.get("enforce_detection", True),
-        align=input_args.get("align", True),
-        expand_percentage=input_args.get("expand_percentage", 0),
+        enforce_detection=input_args.get("enforce_detection", True) if type(input_args.get("enforce_detection", True)) is bool else input_args.get("enforce_detection") == "true",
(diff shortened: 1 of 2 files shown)


deepface/api/src/modules/core/routes.py
```diff
@@ -160,6 +166,8 @@ def analyze():
         detector_backend=input_args.get("detector_backend", "opencv"),
         enforce_detection=input_args.get("enforce_detection", True),
         align=input_args.get("align", True),
+        expand_percentage=input_args.get("expand_percentage", 0),
+        silent=input_args.get("silent", False),
```
serengil, line 170: this is getting silent as string instead of boolean

deepface/api/src/modules/core/routes.py
```diff
@@ -87,6 +87,8 @@ def represent():
         detector_backend=input_args.get("detector_backend", "opencv"),
         enforce_detection=input_args.get("enforce_detection", True),
         align=input_args.get("align", True),
+        expand_percentage=input_args.get("expand_percentage", 0),
```
serengil, line 90: this is getting expand_percentage as string from env vars, you should cast it to int

deepface/api/src/modules/core/routes.py
```diff
@@ -118,8 +120,12 @@ def verify():
         model_name=input_args.get("model_name", "VGG-Face"),
         detector_backend=input_args.get("detector_backend", "opencv"),
         distance_metric=input_args.get("distance_metric", "cosine"),
-        align=input_args.get("align", True),
         enforce_detection=input_args.get("enforce_detection", True),
+        align=input_args.get("align", True),
```
serengil, line 124: same here

deepface/api/src/modules/core/routes.py
```diff
@@ -118,8 +120,12 @@ def verify():
...
+        expand_percentage=input_args.get("expand_percentage", 0),
```
serengil, line 125: same here

deepface/api/src/modules/core/routes.py
```diff
@@ -118,8 +120,12 @@ def verify():
...
+        normalization=input_args.get("normalization", "base"),
+        silent=input_args.get("silent", False),
```
serengil, line 127: same here

deepface/api/src/modules/core/routes.py
```diff
@@ -118,8 +120,12 @@ def verify():
...
+        threshold=input_args.get("threshold", None),
```
serengil, line 128: same here

deepface/api/src/modules/core/routes.py
```diff
@@ -160,6 +166,8 @@ def analyze():
         detector_backend=input_args.get("detector_backend", "opencv"),
         enforce_detection=input_args.get("enforce_detection", True),
         align=input_args.get("align", True),
+        expand_percentage=input_args.get("expand_percentage", 0),
```
serengil, line 169: same here

deepface/api/src/modules/core/routes.py
```diff
@@ -85,10 +85,24 @@ def represent():
         img_path=img,
         model_name=input_args.get("model_name", "VGG-Face"),
         detector_backend=input_args.get("detector_backend", "opencv"),
-        enforce_detection=input_args.get("enforce_detection", True),
-        align=input_args.get("align", True),
-        anti_spoofing=input_args.get("anti_spoofing", False),
-        max_faces=input_args.get("max_faces"),
+        enforce_detection=
```
serengil, line 88: this is not clean at all, for boolean arguments you can adopt this approach

```python
enforce_detection = input_args.get("enforce_detection", "true") == "true"
```

deepface/api/src/modules/core/routes.py
```diff
@@ -85,10 +85,24 @@ def represent():
...
+            bool(input_args.get("align", True))
+            if isinstance((input_args.get("align", True)), bool)
+            else input_args.get("align") == "true",
+        expand_percentage=int(input_args.get("expand_percentage", 0)),
+        normalization=input_args.get("normalization", "base"),
+        anti_spoofing=
+            bool(input_args.get("anti_spoofing", False))
+            if isinstance((input_args.get("anti_spoofing", False)), bool)
+            else input_args.get("anti_spoofing") == "true",
+        max_faces=
```
serengil, line 102: for float args, you can adopt this approach

```python
max_faces = int(input_args.get("max_faces", "0"))
```

deepface/api/src/modules/core/routes.py
charsleysa: Your suggestion will not work for JSON because booleans in JSON are converted to Python booleans, not Python strings

deepface/api/src/modules/core/routes.py
charsleysa: Your suggestion will not work for JSON because numbers in JSON are converted to Python numbers, not Python strings

deepface/api/src/modules/core/routes.py
serengil: Still, there are redudants in your approcah. This would be better imo. 

```python
enforce_detection = input_args.get("enforce_detection", True)
if not isinstance(enforce_detection, bool):
    enforce_detection = enforce_detection == "true"
```

Issue #1380, opened 2024-11-04
[BUG]: Error during DeepFace analysis: The layer sequential_1 has never been called and thus has no defined input.


tora-golia:
### Before You Report a Bug, Please Confirm You Have Done The Following...

- [X] I have updated to the latest version of the packages.
- [X] I have searched for both [existing issues](https://github.com/serengil/deepface/issues) and [closed issues](https://github.com/serengil/deepface/issues?q=is%3Aissue+is%3Aclosed) and found none that matched my issue.

### DeepFace's version

0.0.93

### Python version

3.12.4

### Operating System

macOS 14.6.1

### Dependencies

absl-py==2.1.0
annotated-types==0.7.0
anyio==4.4.0
astor==0.8.1
astunparse==1.6.3
beautifulsoup4==4.12.3
blinker==1.8.2
certifi==2024.8.30
charset-normalizer==3.3.2
click==8.1.7
contourpy==1.3.0
cycler==0.12.1
Cython==3.0.11
decorator==4.4.2
deepface==0.0.93
distro==1.9.0
dlib==19.24.6
face-recognition==1.3.0
face_recognition_models==0.3.0
facenet-pytorch==2.6.0
fastapi==0.115.0
fer==22.5.1
ffmpeg==1.4
filelock==3.16.1
fire==0.6.0
Flask==3.0.3
Flask-Cors==5.0.0
flatbuffers==24.3.25
fonttools==4.53.1
fsspec==2024.9.0
gast==0.6.0
gdown==5.2.0
google-pasta==0.2.0
grpcio==1.66.1
gunicorn==23.0.0
h11==0.14.0
h5py==3.11.0
httpcore==1.0.5
httpx==0.27.2
idna==3.10
imageio==2.35.1
imageio-ffmpeg==0.5.1
imgaug==0.4.0
imutils==0.5.4
itsdangerous==2.2.0
Jinja2==3.1.4
jiter==0.6.1
joblib==1.4.2
keras==3.5.0
kiwisolver==1.4.7
lazy_loader==0.4
libclang==18.1.1
lmdb==1.5.1
lxml==5.3.0
lz4==4.3.3
Markdown==3.7
markdown-it-py==3.0.0
MarkupSafe==2.1.5
matplotlib==3.9.2
mdurl==0.1.2
ml-dtypes==0.4.1
moviepy==1.0.3
mpmath==1.3.0
mtcnn==1.0.0
namex==0.0.8
networkx==3.3
[...]

PR #1456, opened 2025-03-30
Update VGGFace.py

timselier:
Tensorflow versions > 2.13 give error: AttributeError: The layer sequential has never been called and thus has no defined input.

This is solved by replacing `model.input` by `model.layers[0].input`

## Tickets

https://github.com/serengil/deepface/issues/1380

## How to test

```shell
make lint && make test
```


serengil:
Very good spot. Cheers!

Issue #1451, opened 2025-03-10
[FEATURE]: Add Angular Distance as a Distance Metric


serengil:
### Description

DeepFace currently supports cosine distance, Euclidean distance, and Euclidean L2 distance for face embedding comparisons. To enhance distance metric options, we should add angular distance, which is based on the angle between two embeddings. This metric is particularly useful for spherical embeddings and provides an alternative to cosine distance.

# Formula

```python
angular_distance = np.arccos(similarity) / math.pi
```

denizomer:
thank you.

haddyadnan:
Hi @serengil, I implemented this, but after testing it has a coverage of 97.0, we might just need to take a look at the thresholds.

Here is the current threshold:

 "VGG-Face": {
            "cosine": 0.68,
            "euclidean": 1.17,
            "euclidean_l2": 1.17,
            "angular": 0.43,
        },  # 4096d - tuned with LFW
        "Facenet": {"cosine": 0.40, "euclidean": 10, "euclidean_l2": 0.80, "angular": 0.47},
        "Facenet512": {"cosine": 0.30, "euclidean": 23.56, "euclidean_l2": 1.04, "angular": 0.49},
        "ArcFace": {"cosine": 0.68, "euclidean": 4.15, "euclidean_l2": 1.13, "angular": 0.43},
        "Dlib": {"cosine": 0.07, "euclidean": 0.6, "euclidean_l2": 0.4, "angular": 0.50},
        "SFace": {"cosine": 0.593, "euclidean": 10.734, "euclidean_l2": 1.055, "angular": 0.445},
        "OpenFace": {"cosine": 0.10, "euclidean": 0.55, "euclidean_l2": 0.55, "angular": 0.50},
        "DeepFace": {"cosine": 0.23, "euclidean": 64, "euclidean_l2": 0.64, "angular": 0.49},
        "DeepID": {"cosine": 0.015, "euclidean": 45, "euclidean_l2": 0.17, "angular": 0.50},
        "GhostFaceNet": {"cosine": 0.65, "euclidean": 35.71, "euclidean_l2": 1.10, "angular": 0.43},
        "Buffalo_L": {"cosine": 0.55, "euclidean": 0.6, "euclidean_l2": 1.1, "angular": 0.45},
    }

The angular thresholds are based on the cosine value (perhaps that might not be best?)

`np.arccos(1 - np.cos(sim)) / np.pi`

Here are the results (only the ones missed)
<head></head>
[...]

PR #1459, opened 2025-04-07
feature: Add Angular Distance as a Distance Metric

haddyadnan:
## Tickets

https://github.com/serengil/deepface/issues/1451

### What has been done

feature: Added Angular Distance as a Distance Metric
## How to test

```shell
make lint && make test
```

You'll notice the calculation for cosine similarity was repeated rather than calling the find_cosine_distance function. That's because the function includes a 1 - cos(sim) and we can't say for sure that in all cases we'll be able to get the exact values if we do a +1 because calculating the arccos. hence it's safer to just rewrite the calculation excluding the subtraction of 1

### DIFF
--- a/deepface/modules/streaming.py
+++ b/deepface/modules/streaming.py
@@ -51,7 +51,7 @@ def analysis(
             'centerface' or 'skip' (default is opencv).
 
         distance_metric (string): Metric for measuring similarity. Options: 'cosine',
-            'euclidean', 'euclidean_l2' (default is cosine).
+            'euclidean', 'euclidean_l2', 'angular' (default is cosine).
 
         enable_face_analysis (bool): Flag to enable face analysis (default is True).
 
@@ -223,7 +223,7 @@ def search_identity(
             'mtcnn', 'ssd', 'dlib', 'mediapipe', 'yolov8', 'yolov11n', 'yolov11s', 'yolov11m',
             'centerface' or 'skip' (default is opencv).
         distance_metric (string): Metric for measuring similarity. Options: 'cosine',
-            'euclidean', 'euclidean_l2' (default is cosine).
+            'euclidean', 'euclidean_l2', angular, (default is cosine).
     Returns:
         result (tuple): result consisting of following objects
             identified image path (str)
@@ -474,7 +474,7 @@ def perform_facial_recognition(
             'mtcnn', 'ssd', 'dlib', 'mediapipe', 'yolov8', 'yolov11n', 'yolov11s',
             'yolov11m', 'centerface' or 'skip' (default is opencv).
         distance_metric (string): Metric for measuring similarity. Options: 'cosine',
-            'euclidean', 'euclidean_l2' (default is cosine).
+            'euclidean', 'euclidean_l2', angular (default is cosine).
         model_name (str): Model for face recognition. Options: VGG-Face, Facenet, Facenet512,
             OpenFace, DeepFace, DeepID, Dlib, ArcFace, SFace and GhostFaceNet (default is VGG-Face).
     Returns:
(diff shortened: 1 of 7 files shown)


serengil:
`You'll notice the calculation for cosine similarity was repeated rather than calling the find_cosine_distance function. That's because the function includes a 1 - cos(sim) and we can't say for sure that in all cases we'll be able to get the exact values if we do a +1 because calculating the arccos. hence it's safer to just rewrite the calculation excluding the subtraction of 1`

Agree having dedicated function for find_angular_distance is better than putting it into find_cosine_distance. Otherwise, it will be much more complex.

deepface/modules/verification.py
```diff
@@ -51,7 +51,7 @@ def verify(
             'centerface' or 'skip' (default is opencv)
 
         distance_metric (string): Metric for measuring similarity. Options: 'cosine',
-            'euclidean', 'euclidean_l2' (default is cosine).
+            'euclidean', 'euclidean_l2', angular (default is cosine).
```
serengil, line 54: write it with single quote please

deepface/modules/streaming.py
```diff
@@ -223,7 +223,7 @@ def search_identity(
             'mtcnn', 'ssd', 'dlib', 'mediapipe', 'yolov8', 'yolov11n', 'yolov11s', 'yolov11m',
             'centerface' or 'skip' (default is opencv).
         distance_metric (string): Metric for measuring similarity. Options: 'cosine',
-            'euclidean', 'euclidean_l2' (default is cosine).
+            'euclidean', 'euclidean_l2', angular, (default is cosine).
```
serengil, line 226: single quote please

deepface/modules/streaming.py
```diff
@@ -474,7 +474,7 @@ def perform_facial_recognition(
             'mtcnn', 'ssd', 'dlib', 'mediapipe', 'yolov8', 'yolov11n', 'yolov11s',
             'yolov11m', 'centerface' or 'skip' (default is opencv).
         distance_metric (string): Metric for measuring similarity. Options: 'cosine',
-            'euclidean', 'euclidean_l2' (default is cosine).
+            'euclidean', 'euclidean_l2', angular (default is cosine).
```
serengil, line 477: single quote please

serengil (changes requested):
Recommended new thresholds values and some minor changes for typos

deepface/modules/verification.py
```diff
@@ -399,31 +440,32 @@ def find_threshold(model_name: str, distance_metric: str) -> float:
...
         threshold (float): threshold value for that model name and distance metric
             pair. Distances less than this threshold will be classified same person.
     """
 
-    base_threshold = {"cosine": 0.40, "euclidean": 0.55, "euclidean_l2": 0.75}
+    base_threshold = {"cosine": 0.40, "euclidean": 0.55, "euclidean_l2": 0.75, "angular": 0.37}
 
     thresholds = {
-        # "VGG-Face": {"cosine": 0.40, "euclidean": 0.60, "euclidean_l2": 0.86}, # 2622d
+        # "VGG-Face": {"cosine": 0.40, "euclidean": 0.60, "euclidean_l2": 0.86, "angular": 0.37}, # 2622d
```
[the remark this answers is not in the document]
haddyadnan: Buffalo_L should be left as is?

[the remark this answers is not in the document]
serengil: btw, feel free to use 2 digits after point similar to others.

serengil:
@haddyadnan failed because of linting

`deepface/modules/verification.py:452:0: C0301: Line too long (105/100) (line-too-long)`

serengil (approved):
LGTM

PR #1474, opened 2025-05-11
Resolve warnings of Logger library

emmanuel-ferdman:
### What has been done
This PR resolves the deprecation warnings of the `logger` library:
```python
DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
```
It also fixes a few typos along the way.


serengil:
Logger class doesn't have a warning method, it is warn - https://github.com/serengil/deepface/blob/master/deepface/commons/logger.py#L44

How did you decide that it is deprecated?

PR #1479, opened 2025-06-18
Diagram visualization of the deepface codebase

ivanmilevtues:
### What has been done

With this PR, I intorduce high-level diagrams of the deepface codebase to help new contributors orient themselves in the codebase.
You can see how the proposed changes render here:
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/deepface/on_boarding.md

I can see that the deepface has over 2.5K forks. That means that a lot of people are playing around with the codebase. We believe that most of them are interested in changes in just one component of the codebase. This is why we believe diagram first documentation is the way to go, as a person can immediately recognize and navigate to the component of interest. Futher now when they edit it, they have the bigger picture of how this component interacts with the rest of the codebase. Let me know what is your opinion on diagram first documentation!

I'd usually open a discussion first, but you don't have them enabled for this repo so I decided to go ahead and open a PR.

Any feedback is more than welcome!

Full transparency: we’re exploring this idea as a potential startup, but we’re still early and figuring out what’s actually useful to developers.

## How to test

Not relevant

```shell
make lint && make test
```

### DIFF
--- a/.codeboarding/External_Integration_Layer.md
+++ b/.codeboarding/External_Integration_Layer.md
@@ -1,0 +1,163 @@
+```mermaid
+
+graph LR
+
+    API_Core["API Core"]
+
+    Streaming_Module["Streaming Module"]
+
+    DeepFace_Facade["DeepFace Facade"]
+
+    Common_Utilities["Common Utilities"]
+
+    Models["Models"]
+
+    API_Core -- "delegates requests to" --> DeepFace_Facade
+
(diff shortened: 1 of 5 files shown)


ivanmilevtues:
Hi @serengil!
We've introduced a github action which can keep the docs up-to-date. I will setup a quick workflow for you in a new PR :).
It is completely free.

Otherwise we have this demo page which can generate diagrams: https://www.codeboarding.org/demo
But that page will do complete regenerations, while the action will only do updates :)

Let me know if this helps!

PR #1480, opened 2025-06-19
Fix unnecessary catch masking real error

danielgoodwin97:
### What has been done

Remove a try catch which was preventing seeing the real and helpful error message.

Before this change:
```
Exception while processing img2_path
```

After this change:
```
Face could not be detected in images/test.jpg.Please confirm that the picture is a face photo or consider to set enforce_detection param to False.
```

### DIFF
--- a/deepface/modules/verification.py
+++ b/deepface/modules/verification.py
@@ -164,19 +164,16 @@ def verify(
             img_embeddings = [img_path]
             img_facial_areas = [no_facial_area]
         else:
-            try:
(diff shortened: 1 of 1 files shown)


serengil:
Original error was also thrown

```python
raise ValueError(f"Exception while processing img{index}_path") from err
```

Also, it was easy to understand that img1_path or img2_path cause this error. In case you are sending to numpy arrays to img1 and img2, then your approach will offer bad experience.

PR #1482, opened 2025-06-25
Visualization of the deepface codebase. Action to auto-update via manual trigger.

ivanmilevtues:
## Tickets

### What has been done

In this PR, I've added a new intial generation alongside a workflow which when manually triggered will update the docuemntation.

The default values of the manual trigger should be good enough, the result of the workflow is a PR, so that you can check the changes before merging.

One thing you have to setup on yourside for the action to work is:
You need an Action Secret so that the PR workflow part can create the PR.
Settings -> Secrets and variables -> Repository Secret:
The secret has to be named PAT_TOKEN.

These are the steps to generate it:
o to your GitHub account.
Navigate to Settings > Developer settings > Personal access tokens.

Choose either:
Tokens (classic) — easier but less secure.
Fine-grained tokens — more secure and recommended.

For simplicity, use classic token unless you're required to use fine-grained.
Click "Generate new token":
Name: e.g. create-pr-token
Expiration: Set appropriately (e.g., 90 days or No expiration)
Scopes required:
repo → Full access to private/public repositories (Can be also just for the deep face repo)
workflow → To allow the token to trigger other workflows (if needed)

## How to test

Go to actions, on the left there will be the CodeBoarding Documentation update workflow. Run-Workflow and just click run  :) I am here, happy to collaborate!


serengil:
Will be better to generate docs manually

PR #1492, opened 2025-07-22
Pass test suite with python 3.8

ddaanet:
Fix the test suite to pass with python 3.8

Include additional constraints to requirements.txt to avoid incompatible recent versions.

```shell
uv venv --python 3.8
uv pip install -r requirements.txt -r requirements-dev.txt
source .venv/bin/activate
make test
```


ddaanet:
https://github.com/serengil/deepface/issues/1493

serengil:
Would you please also add non-existing path cases for represent and analyze?

In previous version, these were done. You added existing but not valid cases, but we lost non existing cases.

serengil (approved):
LGTM

PR #1499, opened 2025-08-01
Fix FastMtCnn detection error

nico-hahn:
### What has been done
Calling FastMtCnn detection throws an error due to facenet-pytorch only accepting uint8. This PR adds an explicit cast before calling the model to fix that error.

## How to test
* Install dependencies required for FastMtCnn
* Run FastMtCnn detection

### DIFF
--- a/deepface/models/face_detection/FastMtCnn.py
+++ b/deepface/models/face_detection/FastMtCnn.py
@@ -30,6 +30,7 @@ class FastMtCnnClient(Detector):
         resp = []
 
         img_rgb = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)  # mtcnn expects RGB but OpenCV read BGR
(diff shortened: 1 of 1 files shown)


serengil:
I can run fast mtcnn without casting images to int.

torch -> 2.1.1
facenet-pytorch -> 2.5.3

Still, casting it to uint8 doesn't cause any trouble. 

Would you please share your dependency versions and the exception message track.

nico-hahn:
I don't know, I'm still not quite sure what caused the issue for me initially. (I ran the container for quite some time, testing different stuff on my end, with multiple restarts as well.) But when I tried to reproduce it, I basically always got the situation I described earlier.
However, if there's no downside to that cast, maybe it would still make sense to merge it. Dependency management can be weird sometimes, and after all, it would still prevent that error I received, probably regardless of the exact circumstances.

Issue #1506, opened 2025-08-17
[BUG]: Getting AttributeError: 'KerasHistory' object has no attribute 'layer' when calling DeepFace.represent with detector_backend="retinaface"


niteshsahujhansi:
### Before You Report a Bug, Please Confirm You Have Done The Following...

- [x] I have updated to the latest version of the packages.
- [x] I have searched for both [existing issues](https://github.com/serengil/deepface/issues) and [closed issues](https://github.com/serengil/deepface/issues?q=is%3Aissue+is%3Aclosed) and found none that matched my issue.

### DeepFace's version

v0.0.95

### Python version

v3.12.3

### Operating System

Ubuntu 24.04.1

### Dependencies

absl-py==2.3.1
anyio==4.10.0
argon2-cffi==25.1.0
argon2-cffi-bindings==25.1.0
arrow==1.3.0
asttokens==3.0.0
astunparse==1.6.3
async-lru==2.0.5
attrs==25.3.0
babel==2.17.0
beautifulsoup4==4.13.4
bleach==6.2.0
blinker==1.9.0
certifi==2025.8.3
cffi==1.17.1
charset-normalizer==3.4.3
click==8.2.1
comm==0.2.3
debugpy==1.8.16
decorator==5.2.1
deepface==0.0.95
defusedxml==0.7.1
executing==2.2.0
fastjsonschema==2.21.2
filelock==3.19.1
fire==0.7.1
Flask==3.1.1
flask-cors==6.0.1
flatbuffers==25.2.10
fqdn==1.5.1
gast==0.6.0
gdown==5.2.0
google-pasta==0.2.0
grpcio==1.74.0
gunicorn==23.0.0
h11==0.16.0
h5py==3.14.0
httpcore==1.0.9
httpx==0.28.1
idna==3.10
ipykernel==6.30.1
ipython==9.4.0
ipython_pygments_lexers==1.1.1
isoduration==20.11.0
itsdangerous==2.2.0
jedi==0.19.2
Jinja2==3.1.6
joblib==1.5.1
json5==0.12.1
jsonpointer==3.0.0
jsonschema==4.25.0
jsonschema-specifications==2025.4.1
jupyter-events==0.12.0
jupyter-lsp==2.2.6
jupyter_client==8.6.3
jupyter_core==5.8.1
jupyter_server==2.16.0
jupyter_server_terminals==0.5.3
jupyterlab==4.4.6
jupyterlab_execute_time==3.2.0
[...]

PR #1507, opened 2025-08-17
replace training.Model with Model in ArcFace

niteshsahujhansi:
[BUG]: Getting AttributeError: 'KerasHistory' object has no attribute 'layer' when calling DeepFace.represent with detector_backend="retinaface"

## Tickets

Resolves https://github.com/serengil/deepface/issues/1506

### What has been done

**from tensorflow.python.keras.engine import training
model = training.Model(img_input, x, name="ResNet34")**

TensorFlow considers anything under tensorflow.python.* a private/internal API. These APIs are not guaranteed to remain stable and may be blocked or unreliable—even if the file exists.

The public, stable interface is tf.keras.models.Model (or similarly, keras.models.Model in standalone Keras)
which is fully supported across all TensorFlow versions (1.x via standalone Keras, and 2.x via tf.keras)

now its should be updated as

**model = Model(img_input, x, name="ResNet34")
and should remove imports
from keras.engine import training
from tensorflow.python.keras.engine import training**

### DIFF
--- a/deepface/models/facial_recognition/ArcFace.py
+++ b/deepface/models/facial_recognition/ArcFace.py
@@ -15,7 +15,6 @@ tf_version = package_utils.get_tf_major_version()
 
 if tf_version == 1:
     from keras.models import Model
-    from keras.engine import training
     from keras.layers import (
         ZeroPadding2D,
         Input,
@@ -29,7 +28,6 @@ if tf_version == 1:
     )
 else:
     from tensorflow.keras.models import Model
-    from tensorflow.python.keras.engine import training
     from tensorflow.keras.layers import (
         ZeroPadding2D,
         Input,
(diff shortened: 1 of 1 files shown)


serengil:
Doesn't this about ArcFace?

serengil:
No, I got it but retinaface is mentioned in title and description instead of arcface.

niteshsahujhansi:
Oh! I only mentioned retinaface in the title/description because that’s where I first encountered the error. With the same setup, it was working fine when I used the opencv detector backend.

Later I realized the difference comes from how the detectors handle preprocessing:

OpenCV backend → returns a plain cropped face image → DeepFace converts to a proper Keras input. ✅

RetinaFace backend → sometimes returns an array that bypasses Keras Input metadata → when ArcFace tries to connect layers, _keras_history.layer is missing. ❌

So the real issue is not retinaface itself, but how we load ArcFace (training.Model vs. Model).

serengil:
Perfect. Thank you for the explanation. I asked because github uses the title while generating the release notes.

serengil:
Thank you for your contribution, really grateful

PR #1596, opened 2026-04-24
[FEATURE]: delete by img name

sogeking7:
## Overview
This PR adds functionality to delete face embeddings by image name, enabling selective cleanup of the embedding database.

## Changes
- **Core Feature:** New `delete_by_img_name()` function supporting postgres, mongo, weaviate, neo4j, pgvector, and pinecone backends
- **API:** New `DELETE /faces/<img_name>` endpoint with authentication and comprehensive error handling
- **Infrastructure:** Docker cache optimization, improved container naming, and clarified service configuration
- **Dependencies:** Relaxed numpy and psycopg version pinning to use latest compatible versions

## Testing
```shell
make lint && make test

### CODE REVIEW

File: requirements_local [hunk 1]
```diff
@@ -1,4 +1,4 @@
-numpy==1.22.3
+numpy
```
serengil: please don't change anything in this file. this file is to enforce versions of dockerfile. as is, it will install the latest which may be incompatible with other dependencies.

Dockerfile
```diff
@@ -39,15 +39,18 @@ COPY ./entrypoint.sh /app/deepface/api/src/entrypoint.sh
 # RUN pip install --trusted-host pypi.org --trusted-host pypi.python.org --trusted-host=files.pythonhosted.org tensorflow-gpu
 
 # if you plan to use face anti-spoofing, then activate this line
-# RUN pip install --trusted-host pypi.org --trusted-host pypi.python.org --trusted-host=files.pythonhosted.org torch==2.1.2
+RUN --mount=type=cache,target=/root/.cache/pip \
```
serengil: please don't change dockerfile as well

docker/docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 services:
   postgres:
     image: postgres:15
-    container_name: postgres
+    container_name: deepface-postgres
```
serengil, line 4: this should be rolled-back

scripts/dockerize.sh
```diff
@@ -27,10 +27,10 @@ docker build -t deepface .
 ENV_FILE="deepface/api/.env"
 if [ -f "$ENV_FILE" ]; then
     echo ".env found, sending to container"
-    docker run -p 5005:5000 --env-file "$ENV_FILE" deepface
+    docker run --name deepface -p 5005:5000 -v "${HOME}/.deepface/weights:/root/.deepface/weights" --env-file "$ENV_FILE" deepface
```
serengil: same here, this should be rolled-back

deepface/modules/datastore.py
```diff
@@ -625,6 +625,49 @@ def build_index(
...
+            - deleted (int): Number of embeddings removed from the database.
+            - img_name (str): The image name identifier that was deleted.
+    """
+    db_client = __connect_database(
+        database_type=database_type,
+        connection_details=connection_details,
+        connection=connection,
+    )
+
+    deleted = db_client.delete_by_img_name(img_name=img_name)
```
serengil, line 663: delete_by_img_name is only written for postgres. if you are not going to implement for them, then we should define this method in deepface/modules/database/types.py::Database and raise unimplemented error for others.

deepface/modules/datastore.py
```diff
@@ -625,6 +625,49 @@ def build_index(
         os.remove(index_path)
 
 
+def delete_by_img_name(
```
serengil, line 628: instead of delete, we should have unregister function and it should accept img_name or img_id.

deepface/api/src/modules/core/routes.py
```diff
@@ -302,3 +302,28 @@ def build_index() -> Tuple[Dict[str, Any], int]:
         database_type=variables.database_type,
         connection_details=variables.conection_details,
     )
+
+
+
+@blueprint.route("/faces/<img_name>", methods=["DELETE"])
```
serengil, line 359: endpoint should be unregister. it should accept img_name or img_id. so, they should be given in body instead of endpoint url.

deepface/modules/datastore.py
```diff
@@ -625,6 +625,49 @@ def build_index(
...
+    img_name: str,
```
serengil, line 629: img_name is not unique in the db. so, we should unregister only by id maybe.

serengil (changes requested):

deepface/modules/database/postgres.py
```diff
@@ -139,6 +139,47 @@ def delete_by_img_name(self, img_name: str) -> int:
         self.conn.commit()
         return int(deleted or 0)
 
+    def fetch_embedding_by_img_name(
+        self,
+        img_name: str,
```
serengil, line 144: img_name is not unique, this should be done with id.

deepface/modules/datastore.py
```diff
@@ -668,6 +668,115 @@ def delete_by_img_name(
...
+    threshold = find_threshold(model_name=model_name, distance_metric=distance_metric)
+
+    db_client = __connect_database(
+        database_type=database_type,
+        connection_details=connection_details,
+        connection=connection,
+    )
+
+    try:
+        cached = db_client.fetch_embedding_by_img_name(
```
serengil, line 711: if this method won't be defined for other backends, then it should be defined in the database class and raised an unimplemented error.

serengil:
closed due to inactivity

PR #1606, opened 2026-05-13
feat: Add backend-agnostic identity verification architecture

Dev0907:
This PR evolves DeepFace from:
- a local verification + vector search framework

into:
- a scalable biometric authentication platform

by introducing:
- pluggable verification backends
- identity-centric verification workflows
- backend-aware authentication APIs
- enterprise-grade verification architecture

while preserving complete backward compatibility with existing DeepFace behavior.


deepface/DeepFace.py
```diff
@@ -175,6 +176,138 @@ def verify(
...
+            anti_spoofing=anti_spoofing,
+            **kwargs,
+        )
+    finally:
+        backend.close()
+
+    return result
+
+
+def register_identity(
```
serengil: We already have register function, what is the purpose of this?

deepface/modules/vectorstores/base.py
```diff
@@ -0,0 +1,196 @@
...
+from typing import List, Dict, Any, Optional, Union
+import numpy as np
+
+# project dependencies
+from deepface.commons.logger import Logger
+
+logger = Logger()
+
+
+class VectorStore(ABC):
```
serengil: We already have database abstract class

serengil (changes requested):

deepface/DeepFace.py
Dev0907: Yes, I just mixed my the vector embeddings register with this, hence will correct this.

deepface/modules/vectorstores/base.py
Dev0907: Noted.

deepface/modules/database/inventory.py
```diff
@@ -9,6 +9,8 @@
 from deepface.modules.database.weaviate import WeaviateClient
 from deepface.modules.database.neo4j import Neo4jClient
 from deepface.modules.database.pinecone import PineconeClient
+from deepface.modules.database.local import LocalDatabase
+from deepface.modules.database.milvus import MilvusClientDB
```
serengil, line 13: milvus backend is about another PR, shouldn't be in this.

deepface/DeepFace.py
```diff
@@ -175,6 +176,83 @@ def verify(
     )
 
 
+def verify_identity(
```
serengil, line 179: we decided to name this functionality as identify

deepface/DeepFace.py
```diff
@@ -175,6 +176,83 @@ def verify(
...
+    if connection_details is not None:
+        backend_config['store_config'] = (
+            connection_details if isinstance(connection_details, dict)
+            else {'uri': connection_details}
+        )
+    backend = create_verification_backend('vector', backend_config)
+    backend.connect()
+
+    try:
+        result = backend.verify(
```
serengil, line 236: similar to register and search, this function should call datastore.something

also, we are not putting any logic in DeepFace.py, check other functions

deepface/modules/database/local.py
```diff
@@ -0,0 +1,267 @@
...
+
+# project dependencies
+from deepface.modules.database.types import Database
+from deepface.modules.verification import find_distance
+from deepface.commons.logger import Logger
+
+logger = Logger()
+
+
+class LocalDatabase(Database):
```
serengil, line 13: what is the purpose of localdatabase? imo, we don't need this.

deepface/modules/verification_backends/base.py
```diff
@@ -0,0 +1,187 @@
...
+from typing import Dict, Any, Optional, List, Union
+import numpy as np
+
+# project dependencies
+from deepface.commons.logger import Logger
+
+logger = Logger()
+
+
+class VerificationBackend(ABC):
```
serengil, line 12: you don't need any new abstract class for this development. you should use existing database abstract class.

serengil:
thank you for your contribution, i reviewed this in 2nd round, and it seems this requires a lot of work to do. maybe i should implement identify functionality to spend less time.

PR #1612, opened 2026-06-02
docs: README model summary

nizarbadriali:
Here is what has been done,

I've added a brief and concise model comparison summary to the README in order to to help new users understand the pros and cons of the available face recognition models. I compared them based on accuracy, speed, size. This will hopefully make it easier for beginners to choose a model without digging through benchmarks.


serengil:
this was available in the read me and i removed it to keep read me clear. it is also available in benchmarks page. so, no need to add this into read me.

PR #1617, opened 2026-08-02
Use https for the two dlib model download URLs

mohammedix88:
The two dlib model files are fetched over cleartext `http://`, while every other model URL in the project already uses HTTPS. This switches them to match.

```diff
- WEIGHT_URL  = "http://dlib.net/files/dlib_face_recognition_resnet_model_v1.dat.bz2"
+ WEIGHT_URL  = "https://dlib.net/files/dlib_face_recognition_resnet_model_v1.dat.bz2"

- WEIGHTS_URL = "http://dlib.net/files/shape_predictor_5_face_landmarks.dat.bz2"
+ WEIGHTS_URL = "https://dlib.net/files/shape_predictor_5_face_landmarks.dat.bz2"
```

**HTTPS is verified working on the same host and the same files** — I loaded `https://dlib.net/files/` in a browser with no certificate warning, and a HEAD request to `https://dlib.net/files/shape_predictor_5_face_landmarks.dat.bz2` returns `200` with `content-length: 5706710`. Same bytes, same paths, just the scheme.

Why it's worth doing: `download_weights_if_necessary()` doesn't verify a hash or signature after download, so cleartext transport is the only thing standing between the fetch and whatever comes back. For a face-recognition model that difference matters more than usual — the weights determine what `verify()` considers a match.

Two additional suggestions, happy to open them separately or not at all:

- Pinning expected SHA256 digests in `download_weights_if_necessary()` and checking them post-download would also cover the case where the origin itself serves something unexpected, which HTTPS alone doesn't.
- I have a fuller write-up of the reasoning that I'd rather not put in a public PR. Private vulnerability reporting isn't currently enabled on the repo (`/security/advisories/new` returns 404) — if you turn it on I'll send it through properly. Opening a separate issue about that so this PR stays a simple transport change.

— Mohammad Adnan (CyStack)


mohammedix88:
One follow-up that I'd intended to raise as a separate issue — blank issues are disabled on the repo, so I'll put it here rather than misuse a bug template.

**Could you enable private vulnerability reporting?** `Settings → Advanced Security → Private vulnerability reporting → Enable`. It takes a few seconds.

I have a fuller security write-up related to this area that I'd rather not post publicly, and right now there's no way to send it privately — `/security/advisories/new` returns 404 and there's no `SECURITY.md` contact. That leaves a researcher choosing between staying quiet and opening a public issue, and neither is good for your users.

It matters more than usual for a project this size: deepface records roughly **550,000 installs a month** and gets used in identity and verification workflows. Without an advisory there's no CVE, no OSV entry, and no dependency-scanner alert — so today there's no mechanism to notify those users of anything at all.

**Happy to do the work.** If you enable it I'll file a complete report through it — technical detail, CVSS vector, affected versions, references — written so you can just review and publish. I'll also be explicit about what I have and haven't verified, so nothing is taken on trust.

This PR stands on its own either way; it's a plain transport change and doesn't depend on the above.

— Mohammad Adnan (CyStack)

serengil (approved):

PR #1624, opened 2026-09-13
pytorch backend support added

### LINKED ISSUE #1512: Proposal: PyTorch backend support
### Description

Hi,  

DeepFace is tied to TensorFlow/Keras right now. For PyTorch-based projects this is a bit heavy since it means pulling in TF just for DeepFace, which can cause conflicts.  

I’m considering working on a PyTorch backend:  
- re-implement recognition models (ArcFace, VGG-Face, Facenet, etc.) in PyTorch  
- keep the same API (`verify`, `analyze`) so users can switch between TF and Torch  

Would you be open to this idea?  
- As a PR to this repo (e.g. `backend="torch"`)  
- Or as a forked repo with clear credits to DeepFace 

Curious what you think.  

### Additional Info

_No response_

serengil:
## Tickets

Resolves https://github.com/serengil/deepface/issues/1512

### What has been done

  DeepFace can now run on PyTorch as an alternative to TensorFlow. Nine models were ported from
  Keras to PyTorch, the anti-spoofing models were ported from PyTorch to TensorFlow, and
  the framework a user has installed no longer dictates whether DeepFace imports at all.

### Backend engine selection

  A new deepface/commons/backend_utils.py decides which engine to use, without importing either
  framework to do it (importlib.util.find_spec only):

  1. DEEPFACE_BACKEND_ENGINE environment variable, if set (tensorflow or pytorch)
  2. otherwise TensorFlow, when installed
  3. otherwise PyTorch, when installed
  4. otherwise a clear ValueError naming both install commands

  Neither framework is imported at global level unless it is the one in use. This required
  removing the eager model imports from deepface/modules/modeling.py — AVAILABLE_MODELS is now a
  registry of dotted paths keyed by engine, resolved through importlib on first use. Models with
  no backend affinity (Dlib, SFace, Buffalo_L, every detector) are registered under an ANY key and
  work on both. The singleton model cache is unchanged.

  Verified after a full pipeline run (verify on 4 models, analyze on all 4 attributes,
  anti-spoofing):

| engine | tensorflow in sys.modules | torch in sys.modules |
| --- | --- | --- |
| **tensorflow** | True | False |
| **pytorch** | False | True |

  Supporting modules were made framework-neutral so they stay importable either way: package_utils
  (TF moved inside functions), weight_utils, models/FacialRecognition.py, models/Demography.py,
  modules/preprocessing.py (keras.img_to_array → np.asarray, which is what it does internally),
  and modules/demography.py — the latter imported the TF demography models purely to read their
  labels, now shared via a neutral models/demography/DemographyUtils.py.

  A side effect worth calling out: retinaface and mtcnn are no longer imported at module level, so
  import deepface no longer pulls TensorFlow in through retina-face. Both now raise an actionable
  message if the package is missing.

  ### Models ported to PyTorch

| model | file | notes |
| --- | --- | --- |
| **VGG-Face** | `facial_recognition/pytorch/VGGFace.py` | fc6/fc7/fc8 kept as convolutions |
| **FaceNet-128d / 512d** | `facial_recognition/pytorch/Facenet.py` | Inception-ResNet-v1, parameterized instead of the 1707-line unrolled graph |
| **ArcFace** | `facial_recognition/pytorch/ArcFace.py` | pre-activation ResNet34 + PReLU |
| **GhostFaceNet** | `facial_recognition/pytorch/GhostFaceNet.py` | ghost bottlenecks + SE |
| **Age / Gender / Race** | `demography/pytorch/` | share a VggFaceClassifier (VGG trunk + 1×1 conv head) |
| **Emotion** | `demography/pytorch/Emotion.py` | standalone CNN on 48×48 grayscale |

  Two shared bases keep the clients drop-in compatible: TorchFacialRecognition and TorchDemography
  accept the same channels-last BGR arrays as the TF clients and transpose internally.

  ### Anti-spoofing: PyTorch → TensorFlow

  FasNet was already PyTorch, so it moved to spoofing/pytorch/ (via git mv) and gained a Keras
  implementation at spoofing/, making anti-spoofing available without torch. Cropping helpers
  shared by both live in spoofing/FasNetUtils.py.

  ### Accuracy verification

  Every port was validated against its original with the other framework as source of truth, over
  12–64 faces from tests/unit/dataset:

 | model | relative drift | decision agreement |
| --- | --- | --- |
| **VGG-Face** | 1.9e-06 | 0/66 verdict changes (euclidean, euclidean_l2, cosine) |
| **FaceNet-128d / 512d** | 1.2e-06 | 0/66 |
| **ArcFace** | 2.3e-06 | 0/66 |
| **GhostFaceNet** | 1.9e-06 | 0/66 |
| **Age** | — | max Δ 9.8e-06 years |
| **Gender** | — | 64/64 labels, max Δp 6.6e-07 |
| **Race** | — | 64/64 labels, max Δp 1.1e-06 |
| **Emotion** | — | 64/64 labels, max Δp 1.6e-06 |
| **FasNet (V2 / V1SE)** | 2.0e-05 | 64/64 is_real |

  Drift is float32 accumulation, not conversion error. The tightest margin anywhere — an ArcFace
  pair sitting 0.0017 from the euclidean threshold — has ~380× headroom over the backend
  difference.

### Packaging
  
  setup.py gains extras_require, fed by two new files:    

  requirements_tf.txt       tensorflow, keras, retina-face, mtcnn
  requirements_pytorch.txt  torch>=2.1.2

  requirements.txt is unchanged for now — pip install deepface still installs TensorFlow, so this
  release is not breaking. requirements_base.txt holds the backend-free base intended for the next
  major release, and DeepFace.py prints a deprecation warning announcing the switch to explicit
  deepface[tensorflow] / deepface[pytorch].

  For now, if one wants to install pytorch only, they should follow these steps

  ```
  $ pip install -r requirements_base.txt && pip install -r requirements_pytorch.txt
  $ pip install deepface --no-deps
  ```

### Other

  - make test-tf / make test-pytorch run the unit suite pinned to each engine
  - New weights (9 .pth, 2 .h5) to be published under [deepface_models](https://github.com/serengil/deepface_models/releases/tag/v1.0).

## How to test

```shell
make lint && make test
```


deepface/DeepFace.py
```diff
@@ -37,16 +35,40 @@
 logger = Logger()
 
 # -----------------------------------
-# configurations for dependencies
+# warn users about upcoming changes in backend installation.
+
+if "_DEPRECATION_WARNING_SHOWN" not in globals():
```
bulentsiyah, line 40: module body only runs once per process, so this flag is never set when the check
runs - logger.warn alone behaves exactly the same. the module level `global` is a
no-op too, that's why pylint had to be silenced here.

Makefile
```diff
@@ -1,6 +1,12 @@
 test:
 	cd tests/unit && python -m pytest . -s --disable-warnings
 
+test-tf:
+	DEEPFACE_BACKEND_ENGINE=tensorflow $(MAKE) test
+
+test-pytorch:
```
bulentsiyah, line 7: tests/unit/test_commons.py calls package_utils.get_tf_major_version() and imports
keras at module level, so this target fails at collection on a torch-only install.
also, the torch load_model_weights has no counterpart of test_loading_broken_weights.

README.md
```diff
@@ -51,15 +51,25 @@ DeepFace is a lightweight [face recognition](https://sefiks.com/2018/08/06/deep-
...
+$ export DEEPFACE_BACKEND_ENGINE=pytorch
 ```
 
 Alternatively, you can also install deepface from its source code. Source code may have new features not published in pip release yet.
 
 ```shell
 $ git clone https://github.com/serengil/deepface.git
 $ cd deepface
-$ pip install -e .
+$ pip install -e .[tensorflow]  # or pip install -e .[pytorch]
```
bulentsiyah, line 72: this one pulls tensorflow in through install_requires as well, same reason the
pypi pytorch line above is commented out. should it be commented too until the
base switch, or point to requirements_base.txt + requirements_pytorch.txt instead?

bulentsiyah (commented):
deciding the backend with find_spec, without importing either framework, is the
right call - and the registry keeps the singleton intact while making imports lazy.
a few things:

nothing in CI runs the torch code. tests.yml isn't touched here and its unit test
step is already stubbed out, and the type-check job does `pip install .` which
brings tensorflow only - with ignore_missing_imports = True that makes torch Any,
which is why all 19 nn.Module subclasses needed `# type: ignore[misc]`. torch ships
py.typed, so running mypy once more with torch installed and tensorflow absent
would both type-check the new modules for real and catch a module level
`import tensorflow` sneaking back in later.

the parity numbers look solid but the scripts behind them aren't in the tree.
worth committing one as a test, skipped unless both frameworks are importable?
otherwise it has to be redone by hand at the next weight bump.

last, a which-model-runs-on-which-engine table in the README would help, since
DeepID/DeepFace/OpenFace stay tf only. get_model_class raises a good error but
people will look for the table first.

serengil:
@bulentsiyah Thanks for the detailed review! A few clarifications on these points:

Mypy & CI: On my local machine with both PyTorch and TF installed, Mypy passed clean without any issues. I only added ⁠# type: ignore[misc]⁠ because CI was failing due to missing PyTorch dependencies. Adding additional pytorch dependency to ci will cause extra load, and I once suspended by github. I don't want this to happen again.

Parity Scripts: The parity verification is just a simple Cosine / Euclidean distance check on the outputs. Since these model weights won't be re-converted, keeping temporary scripts in the repo would just add noise, so I prefer leaving them out.

TF-only Models & README: I'm actually planning to convert ⁠DeepID⁠, ⁠OpenFace⁠, and ⁠DeepFace⁠ to PyTorch tomorrow as well! Once that's done, all models will support both engines.

setup.py
```diff
@@ -35,4 +60,8 @@
     python_requires=">=3.7",
     license="MIT",
     install_requires=requirements,
+    extras_require={
+        "tensorflow": tensorflow_requirements,
+        "pytorch": pytorch_requirements,
```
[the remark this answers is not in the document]
serengil: this is intentional not to have breaking change. currently, we're showing a warning message that tf will no longer be installed by default pip install deepface command.

Makefile
```diff
@@ -1,6 +1,12 @@
...
+	DEEPFACE_BACKEND_ENGINE=pytorch $(MAKE) test
```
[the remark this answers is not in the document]
serengil: unit tests should be run with both tf and pytorch.

requirements_pytorch.txt
```diff
@@ -0,0 +1 @@
+torch>=2.1.2
```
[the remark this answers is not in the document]
serengil: i'm currently using Python 3.8.12, we can switch requires condition with this.

README.md
serengil: we will do it in the next release. if we do this now, then people using `pip install deepface` will be broken. now, we are showing a warning message about this, and with next release we will use requirements_base instead of requirements.

deepface/DeepFace.py
serengil: i tried warning.warn but it didn't show anything in the terminal.

Makefile
serengil: i expect one to have both tf and ptc while running unit tests.

deepface/models/facial_recognition/pytorch/VGGFace.py
```diff
@@ -0,0 +1,204 @@
...
+    Returns:
+        model (VggFaceDescriptor): returning 4096 dimensional vectors
+    """
+    model = base_model()
+
+    weight_file = weight_utils.download_weights_if_necessary(
+        file_name="vgg_face_weights.pth", source_url=url
+    )
+
+    model = load_model_weights(model=model, weight_file=weight_file)
```
fcakyon, line 199: please clear `model.fc8` after checkpoint loading to save 41 MiB of unused FP32 weights, since [the PyTorch descriptor only calls embed()](https://github.com/serengil/deepface/blob/6537af669af824d2bf0e3aa0fa285195b37eb79e/deepface/models/facial_recognition/pytorch/VGGFace.py#L159-L173) while [TensorFlow already excludes the classifier from its returned graph](https://github.com/serengil/deepface/blob/6537af669af824d2bf0e3aa0fa285195b37eb79e/deepface/models/facial_recognition/VGGFace.py#L161-L168)

deepface/models/demography/pytorch/Age.py
```diff
@@ -0,0 +1,98 @@
...
+        age_predictions = self._predict_internal(imgs)
+
+        # Calculate apparent ages
+        if len(age_predictions.shape) == 1:  # Single prediction list
+            return find_apparent_age(age_predictions)
+
+        return np.array([find_apparent_age(age_prediction) for age_prediction in age_predictions])
+
+
+class AgeModel(VggFaceClassifier):
```
fcakyon, line 66: please use `VggFaceClassifier(CLASSES)` directly and delete the trivial subclasses and `base_model()` wrappers in [Age](https://github.com/serengil/deepface/blob/6537af669af824d2bf0e3aa0fa285195b37eb79e/deepface/models/demography/pytorch/Age.py#L66-L98), [Gender](https://github.com/serengil/deepface/blob/6537af669af824d2bf0e3aa0fa285195b37eb79e/deepface/models/demography/pytorch/Gender.py#L60-L92) and [Race](https://github.com/serengil/deepface/blob/6537af669af824d2bf0e3aa0fa285195b37eb79e/deepface/models/demography/pytorch/Race.py#L62-L94) to remove about 51 lines, as [TensorFlow already builds these models directly](https://github.com/serengil/deepface/blob/6537af669af824d2bf0e3aa0fa285195b37eb79e/deepface/models/demography/Gender.py#L67-L88) and its shared [VGGFace.base_model()](https://github.com/serengil/deepface/blob/6537af669af824d2bf0e3aa0fa285195b37eb79e/deepface/models/facial_recognition/VGGFace.py#L83-L134) contains the actual architecture and should stay

deepface/models/face_detection/RetinaFace.py
```diff
@@ -3,17 +3,35 @@
...
 
 # pylint: disable=too-few-public-methods
 class RetinaFaceClient(Detector):
     def __init__(self) -> None:
         """RetinaFace face detector model initialization"""
-        self.model = rf.build_model()
+        self.model = self.import_retinaface().build_model()
+
+    @staticmethod
+    def import_retinaface() -> Any:
```
fcakyon, line 21: please move the retinaface import and dependency error to module scope and delete this helper, because [the shared model registry already loads this module on demand](https://github.com/serengil/deepface/blob/6537af669af824d2bf0e3aa0fa285195b37eb79e/deepface/modules/modeling.py#L152-L154) and [detect_faces calls the extra helper on every invocation](https://github.com/serengil/deepface/blob/6537af669af824d2bf0e3aa0fa285195b37eb79e/deepface/models/face_detection/RetinaFace.py#L48), so one shared fix covers both backends

deepface/commons/backend_utils.py
```diff
@@ -0,0 +1,123 @@
...
+            return _backend_engine
+
+    raise ValueError(
+        "deepface requires either tensorflow or pytorch to be installed, but none of them "
+        "is available. Please run `pip install deepface[tensorflow]` or "
+        "`pip install deepface[pytorch]` to install the backend engine you want to use."
+    )
+
+
+def is_tensorflow() -> bool:
```
fcakyon, line 108: please delete these two unused shared helpers, as a repository-wide search found no callers and [DeepFace](https://github.com/serengil/deepface/blob/6537af669af824d2bf0e3aa0fa285195b37eb79e/deepface/DeepFace.py#L64-L71) and [modeling](https://github.com/serengil/deepface/blob/6537af669af824d2bf0e3aa0fa285195b37eb79e/deepface/modules/modeling.py#L141-L142) already use `get_backend_engine()` directly

deepface/commons/weight_utils.py
```diff
@@ -101,112 +94,76 @@ def load_model_weights(model: Sequential, weight_file: str) -> Sequential:
...
+    ("spoofing", "Fasnet", ["FIRST_WEIGHTS_URL", "SECOND_WEIGHTS_URL"]),
+    # face detection
+    ("face_detector", "ssd", ["MODEL_URL", "WEIGHTS_URL"]),
+    ("face_detector", "yunet", ["WEIGHTS_URL"]),
+    ("face_detector", "dlib", ["WEIGHTS_URL"]),
+    ("face_detector", "centerface", ["WEIGHTS_URL"]),
+]
+
+# models whose weight file is not named after its url
+WEIGHT_FILE_NAMES = {
```
fcakyon, line 126: please move GhostFaceNet's filename beside its URL in [the TensorFlow model module](https://github.com/serengil/deepface/blob/6537af669af824d2bf0e3aa0fa285195b37eb79e/deepface/models/facial_recognition/GhostFaceNet.py#L54-L77) and reuse it in both download paths to remove this shared exception table, following [the existing YOLO filename/URL pattern](https://github.com/serengil/deepface/blob/6537af669af824d2bf0e3aa0fa285195b37eb79e/deepface/commons/weight_utils.py#L164-L166)

